PR #1, opened 2018-12-03
setup travis CI; add Makefile and initial README.md

jhump:
These three files were basically copied from `grpcurl`, but with appropriate edits made. (The README obviously had the most edits. I still want to revise it further and include screenshots of the tool. But I figured this would suffice for now...)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,9 @@ matrix:
       env: VET=1
     - go: "1.11"
       env: GO111MODULE=off
-    - go: "1.11"
-      env: GO111MODULE=on
+# TODO: add go.mod so we can test this way, too
+#    - go: "1.11"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+      env: VET=1
+    - go: "1.11"
+      env: GO111MODULE=off
+# TODO: add go.mod so we can test this way, too
+#    - go: "1.11"
+#      env: GO111MODULE=on
+    - go: tip
+
+script:
+  - if [[ "$VET" = 1 ]]; then make; else make deps test; fi
```
ianrose14: nit: want to do `then make ci` instead of `then make`?  As it is now if you add another make rule before the `ci` rule, everything will change behavior

PR #8, opened 2018-12-10
add CSRF protection to grpcui command-line tool and standalone.Handler

### LINKED ISSUE #6: CSRF protection
The main package needs some way for embedded forms to enable CSRF protection. The `standalone` package (and thus the `grpcui` command-line tool, too) then need to use that mechanism to enable CSRF protection.

Initial idea on how to solve:
1. Provide a `func(r *http.Request) string` function to `grpcui.RPCInvokeHandler`. If supplied, the handler will look for a custom header (`X-CSRF-Token`?) and make sure it matches the string returned by the given function.
2. Update `grpcui.WebFormScript` to expose API (e.g. add exported functions to `window`) that can be used to set a CSRF token. If set, all calls to invocation handler will add it in custom header.
3. In the `standalone` package, use a JS-visible cookie to track the CSRF token. When the main page is fetched, set the cookie if it's not present, generating new tokens using a secure RNG. The custom function given to `grpcui.RPCInvokeHandler` will retrieve the cookie value (and fail with a 401 `Unauthorized` if it does not exist). In the main page, in which the gRPC form is embedded, add a `<script>` tag that queries the cookie value and then calls the script's exported function to set its CSRF token.

jhump:
Fixes #6


internal/resources/standalone/index-template.html
```diff
@@ -97,6 +97,16 @@
       padding: 24px;
     }
     </style>
+    <script>
+    // ZOMG this is awful. This beast comes from https://developer.mozilla.org/en-US/docs/Web/API/document/cookie
```
anwar-fs: Yeah that is monstrous. It seems like if it's consistent enough to write a regex for that, then they could have made a decent programming API.

internal/resources/standalone/index-template.html
```diff
@@ -97,6 +97,16 @@
...
+    var cookieValue = document.cookie.replace(/(?:(?:^|.*;\s*)_grpcui_csrf_token\s*\=\s*([^;]*).*$)|^.*$/, "$1");
+    // add CSRF token to all non-GET requests
+    $(document).ajaxSend(function(event, request, settings) {
+        if (settings.method !== "GET") {
```
anwar-fs: Is there any reason to not include the CSRF token in GET requests? It seems like that wouldn't hurt, and it would provide a level of safety that protects side effect inducing GET requests that get created inadvertently. Also, the wrapper that validates the CSRF token seems to not skip GET requests.

jhump: I guess not. I can have it apply to everything, it will just be ignored.

The thing that validates CSRF is a wrapper that only a protects a single handler, the `grpcui.RPCInvokeHandler`, which only accepts POST requests. (It results in "405 Method Not Allowed" for anything else.)

jhump: FWIW, I formulated it this way so it only adds the header for requests where it will actually get validated. The only non-GET XHRs that this thing does is for invocations. The GET XHRs are for a different handler (one that does not check CSRF token because it isn't necessary -- it's just for querying RPC schema metadata down to the client).

anwar-fs:
LGTM. (The GET request issue is not critical IMO).

ZackFS:
LGTM. GET requests shouldn't :) make any state changes.

Also, what is that humongous hex array?

jhump:
> Also, what is that humongous hex array?

That is a gzip'ed and encoded version of the HTML file. We use `gobindata` to turn static files into compilable code. That way the library has everything it needs without any external dependencies (like files on the file system it must find).

PR #10, opened 2018-12-18
make bind address configurable, add dockerfile

termie:
bind was always 127.0.0.1 which made this very hard to work with from docker and kubernetes, this small change should fix it, added a Dockerfile, too but happy enough to remove it if you don't want it.

### DIFF
--- a/cmd/grpcui/grpcui.go
+++ b/cmd/grpcui/grpcui.go
@@ -88,6 +88,8 @@ var (
 
 	port = flags.Int("port", 0, prettify(`
 		The port on which the web UI is exposed.`))
+	bind = flags.String("bind", "127.0.0.1", prettify(`
+		The address on which the web UI is exposed.`))
 	services multiString
 	methods  multiString
 )
@@ -358,11 +360,11 @@ func main() {
 		})
 	}
 
-	listener, err := net.Listen("tcp", fmt.Sprintf("127.0.0.1:%d", *port))
+	listener, err := net.Listen("tcp", fmt.Sprintf("%s:%d", *bind, *port))
 	if err != nil {
 		fail(err, "Failed to listen on port %d", *port)
 	}
-	fmt.Printf("gRPC Web UI available at http://127.0.0.1:%d/...\n", listener.Addr().(*net.TCPAddr).Port)
+	fmt.Printf("gRPC Web UI available at http://%s:%d/...\n", *bind, listener.Addr().(*net.TCPAddr).Port)
 
 	if err := http.Serve(listener, handler); err != nil {
 		fail(err, "Failed to serve web UI")
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,4 @@
+FROM golang:1.10
+ADD . /go/src/github.com/fullstorydev/grpcui
+WORKDIR /go/src/github.com/fullstorydev/grpcui
+RUN make deps install


cmd/grpcui/grpcui.go
```diff
@@ -358,11 +360,11 @@ func main() {
 		})
 	}
 
-	listener, err := net.Listen("tcp", fmt.Sprintf("127.0.0.1:%d", *port))
+	listener, err := net.Listen("tcp", fmt.Sprintf("%s:%d", *bind, *port))
 	if err != nil {
 		fail(err, "Failed to listen on port %d", *port)
 	}
-	fmt.Printf("gRPC Web UI available at http://127.0.0.1:%d/...\n", listener.Addr().(*net.TCPAddr).Port)
+	fmt.Printf("gRPC Web UI available at http://%s:%d/...\n", *bind, listener.Addr().(*net.TCPAddr).Port)
```
jhump: If the bind address is `0.0.0.0` (bind to all interfaces), this should detect that and show `127.0.0.1` instead. Not sure if there is anything useful in `net` package to help detect this (other than string comparison) -- it would be nice if there were.

rdsubhas: yep, that's nice for usability. but two queries pls:

1. do you think it might be misleading? i.e. if user provides `-bind 0.0.0.0` and sees `available at http://127.0.0.1` wouldn't they be confused as to whether their option was picked up or not?
2. 0.0.0.0 is one case, but there are also [more possibilities](https://golang.org/pkg/net/#SplitHostPort) for binding such as `::` for IPv6, `::1`, `[::1]`, for localhost in IPv6, etc, and `net.Listen` supports all those bind options. but finding the right URL to print for the CLI user in all those cases is pretty tough, e.g. `http://[::1]:8080` would be as obscure as `http://0.0.0.0:8080`

rails/rack/puma/tomcat webservers print whatever the user specified directly ,such as `Listening on <whatever user entered>:<port>`, stepping out of the ambiguity, but I'm not sure if that can apply here...

thanks for the amazing project!!

jhump: > do you think it might be misleading?

Not particularly. The intent is to provide a valid URL, so I'm more concerned about printing something that may not be useful. If you want a separate log line (maybe enabled with `-v`) that prints the bind address separately, that seems fine to eliminate any ambiguity here.

> ... but there are also more possibilities for binding ...

I'm not concerned about obscurity; my concern is just about printing a valid URL. So the address `http://[::1]:8080` is fine, since it would work fine in a browser.

Having said that, it appears that `http://0.0.0.0:8080` and `http://[::]:8080` *do actually work*, at least on my Macbook. What I'm not certain is whether that will work everywhere. My understanding was that, as a host address, 0.0.0.0 just means the default route. Will that always be localhost?

So... In summary, if 0.0.0.0 will always work as a valid URL and route to localhost, then I suppose I withdraw my objection.

rdsubhas: So far tested with curl and wget on the terminal, major mac browsers (chrome, FF, safari) and go `net/http`, java `URLConnection` and node `http.get` code on mac and linux (ubuntu 18.04). It was supported in all those places. unfortunately don't have a canonical link to show it tho, so will ping snippets here...

```
curl --head http://0.0.0.0:8000
HTTP/1.0 200 OK

wget http://0.0.0.0:8000
--2018-12-21 22:04:54--  http://0.0.0.0:8000/
Connecting to 0.0.0.0:8000... connected.
HTTP request sent, awaiting response... 200 OK

# java
new URL("http://0.0.0.0:8000").openConnection().getHeaderFields()

# node
const http = require('http');
http.get('http://0.0.0.0:8000', res => console.log(res));

# go
resp, err := http.Get("http://example.com/")
log.Println(resp)

# and major browsers...
```

I don't have a windows machine to test, but yes know for sure that chrome & firefox URL handling is same across platforms. So for most web development it works out of the box 👍

jhump: Thanks for poking around! Sounds like my one concern is moot.

rdsubhas: Thanks!!

jhump:
Hmmm. I just realized that I didn't want the Dockerfile in here.

For now please maintain yourself, separately. (I'll submit a PR in a sec to delete it.)

We are still evaluating our approach to supporting standard Docker images for our open-source stuff. But the file in here is probably not the way we'd do it since the resulting image is heavy -- it includes the Go SDK and fetched deps. It also does not define an entry point. Any image we provide will likely use a separate build image from base image so that the final image is as light as possible (just the statically linked binary on top of a lightweight base).

PR #20, opened 2019-03-05
Enable automatic reconnection to server

magJ:
Not the cleanest PR, but a much needed feature.

It's really annoying when developing a service, needing to restart grpc ui each time you restart your server app.

I basically just copied the `grpcurl` dial command and removed the connection cancellation logic.

### DIFF
--- a/cmd/grpcui/grpcui.go
+++ b/cmd/grpcui/grpcui.go
@@ -88,6 +88,10 @@ var (
 	veryVeryVerbose = flags.Bool("vvv", false, prettify(`
 		Enable the most verbose output, which includes traces of all HTTP
 		requests and responses.`))
+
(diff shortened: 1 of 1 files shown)


jhump:
@magJ, this is actually fixed in latest master in [grpcurl](https://github.com/fullstorydev/grpcurl/commit/5516a45602fd3ee563a85daaa4ef48fa5a92ffbd#diff-a1f17ecc5f401c05ea33f0dd1c6de9caR576). If you aren't using Go modules, you can just fetch latest version of that dependency.

If you are using Go modules, then I still need to make a release in the `grpcurl` repo and then update this repo's `go.mod` to rely on that new version.

jhump:
I opened an issue to track the state of the current defect: #21

Issue #37, opened 2019-05-24
change the way assets are baked in: go-bindata previously used by go:generate directives no longer available


fnchooft:
Great project!!

Trying to alter some javascript, but it seems i need go-bindata for this.

Is go-bindata still working for this project?
Since i am not a go-expert i am a bit stuck on this for now.
Some GO-expert might add a section in the README on how to re-compile the assets?

Kind regards,

Using latest go version go1.12.5 linux/amd64 on ubuntu 18.04.2

rokostik:
I'm having similar issues, seems like the current versions of `go-bindata` that I found don't accept the same flags and produce a totally different output.

Could anyone point out the actual `go-bindata` that was used in this project?

jhump:
@fnchooft, sorry I never replied. Something must have been wonky w/ my email notifications, because I never saw your original post.

@rokostik, looks like the version of `go-bindata` I used is no longer available on GitHub. It was SHA 4bf44a02f12e6d3506dd4f79bd4d5555ea77c2d1 from `github.com/jteeuwen/go-bindata`. I know this because I still have this tool locally, as well as details of the commit. So I can re-generate stuff in this repo on my machine.

So this will take a change to update this repo to use a properly supported tool, so that contributions can also re-generate.

Happy to accept a patch.

jhump:
Resolved by #49

PR #49, opened 2019-09-21
Fix code to be compatible with newer version of go-bindata (#37)

rokostik:
This PR fixes the code dealing with assets to work with newer versions of go-bindata (I used https://github.com/go-bindata/go-bindata).

Instead of one generated .go file per asset file, it now generates a `bindata.go` file containing all assets. Getting the contents is no longer compile time safe, but is now done with `.MustAsset()` that panics if the asset is not present.


standalone/standalone.go
```diff
@@ -42,34 +42,33 @@ func Handler(ch grpcdynamic.Channel, target string, methods []*desc.MethodDescri
 
 	var mux http.ServeMux
 
-	for resourcePath, val := range standalone.GetResources() {
+	for _, assetName := range standalone.AssetNames() {
+		// the index file will be handled separately
+		if assetName == "index-template.html" {
```
rokostik: Not the prettiest way of handling this, but I think it's not that bad.

rokostik:
Looks like the generated `bindata.go` files are not in line with what `gofmt -s` produces, but regular `gofmt` calls without the `-s` makes no changes.

@jhump do you have any preferences what to do in this case? Should `make checkgofmt` not use the `-s` flag when calling `gofmt` or should the generated files simply be run through `gofmt -s`?

standalone/standalone.go
```diff
@@ -116,7 +115,7 @@ func Handler(ch grpcdynamic.Channel, target string, methods []*desc.MethodDescri
 	})
 }
 
-var indexTemplate = template.Must(template.New("index.html").Parse(string(standalone.GetIndexTemplate())))
+var indexTemplate = template.Must(template.New("index.html").Parse(string(standalone.MustAsset("index-template.html"))))
```
jhump: How about a constant for this name?

webform.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/fullstorydev/grpcui/internal/resources/webform"
 )
 
-var webFormTemplate = template.Must(template.New("grpc web form").Parse(string(webform.WebFormTemplate())))
+var webFormTemplate = template.Must(template.New("grpc web form").Parse(string(webform.MustAsset("webform-template.html"))))
```
jhump: Hmm, that's gross. So the new gobindata doesn't generate functions you can use? It's a shame to lose the type/compile-time safety and have to revert to runtime string checks...

Maybe add provide such functions, hand-written, and add a test that all of the functions return data w/out panic'ing?

webform.go
```diff
@@ -97,13 +97,13 @@ func WebFormContents(invokeURI, metadataURI string, descs []*desc.MethodDescript
 // that first verifies the CSRF token header before delegating to a
 // RPCInvokeHandler.
 func WebFormScript() []byte {
-	return webform.WebFormScript()
+	return webform.MustAsset("webform.js")
```
jhump: Per other comment, we should revert this change and just hand-write `WebFormScript` and company if necessary. This package should not need to know internal details like these file names.

rokostik:
Made changes as per comments.

Not sure if I understand your comment regarding `go generate` though. AFAIK the `generate` command only looks for binaries on the system and has nothing to do with modules. Please correct me if I'm mistaken, since I also think it would be good to have the `go-bindata` version baked in somewhere.

Also, the problem with `gofmt` and `bindata.go` (a few comments above) is still present and the build will fail.

jhump:
@rokostik, if you have modules enabled and you do things like `go get`, the versions of those tools will be recorded. That allows you to use `go get` in CI and still have repeatable builds, because the tools you use are also pinned.

If you do the following, it should add a line to `go.mod` and two lines to `go.sum` with the version info:

```
cd go/src/github.com/fullstorydev/grpcui
GO111MODULE=on go get github.com/a-urth/go-bindata
```

Please also add a comment to `go.mod` explaining that this version is needed for `go generate`.

Regarding the fact that the output is not properly formatted per the CI check, you can fix that by changing the `go-bindata-all.sh` script (and any other `go:generate` directives that run `go-bindata`) to actually run `gofmt -w -s` after the generator program is run.

rokostik:
Thank you very much for the patience and all the advice.

I should be good to go ready now.

PR #54, opened 2019-10-22
Add optional `-prefix` flag to serve UI under path prefix

gfelbing:


jhump:
@gfelbing, why is this useful? Since there is nothing else served by this program, it seems strange to force the path to something else. If you are instead putting it behind a reverse proxy that will serve it via some other path, you should be able to configure the reverse proxy to strip a path prefix, no?

gfelbing:
Hi @jhump ,
we consider to use the UI as a sidecar container within our development cluster for convenient testing of grpc services.
To setup the routing (we are using istio) based on path prefixes, it is easier to have a single prefix for the whole grpcui (static content as well as API).

jhump:
@gfelbing, I think you can configure Istio to route the requests to grpcui, but strip the path prefix. See https://github.com/istio/istio/issues/8076 for a lot of discussion around ways to configure this. (Admittedly, looks like it's less straight-forward than it should be, but it looks do-able.)

gfelbing:
@jhump , unfortunately, URL rewrite does not work here.
The URLs rendered by grpcui do not contain the path prefix if they are stripped by the proxy.
This leads to a broken web UI.

gfelbing:
@jhump ?

jhump:
@gfelbing, sorry for delays. I wanted to investigate some other options before deciding on whether this is the best solution or not. But I've been very busy at work with other things lately, so haven't been able to finish.

jhump:
@gfelbing, the investigation I wanted to do was just to reproduce the issue you mentioned: that the UI generates bad links if it serves everything from "/" and is then put behind a reverse proxy that routes (and strips) some other path prefix to it.

That is what should be fixed. And, upon closer inspection of the pull request, it wouldn't address that anyway. You are just stripping the path from incoming requests and are not trying to rewrite any links. So I am curious how this PR would even fix the issue. If you have encountered this problem, could you let me know exactly what links break? In the meantime, I hope to maybe get to trying to repro this later today or tomorrow. 🤞

gfelbing:
@jhump the PR would fix that issue in a way that one could put grpcui behind a reverse proxy, routing requests by path prefix without rewriting them to '/'

The solution to fix the broken links when rewriting paths would also be nice and actually I love to see both options.

The broken links where mainly css/js sources, as well as the '/invoke' calls. I can reproduce them again on Monday to tell you the exact ones.

jhump:
This PR doesn't seem to do what you say it does. The _only thing_ your change does is strip the prefix from URLs, _exactly_ like a reverse proxy could do. You haven't made any changes to include the path prefix in any of the links. So they would still be written to the response as if based on the root `/` path, not based on the configured path prefix. So things should still be broken.

If things work with this PR, that suggests to me that they should work _without_ this PR, just having the reverse proxy strip the path prefix.

Am I making any sense? I feel a little like we're talking in circles. That's why I wanted to repro myself, to clearly understand what it was you observed. Because things aren't adding up for me.

jhump:
> The URLs rendered by grpcui do not contain the path prefix if they are stripped by the proxy.
This leads to a broken web UI.

I was unable to reproduce. The grpcui HTML only uses relative links, so it works fine if proxy strips the prefix. I just tested it, and everything worked: all assets (CSS, etc) and all XHRs.

I simply setup a simple proxy on port 57633 that strips `/some-prefix/` and forwards to grpcui, and navigating the browser to `http://127.0.0.1:57633/some-prefix/` works just fine. Everything in the UI behaved.

Can you provide more details about the setup you tried where you encountered a broken web UI?

gfelbing:
That sounds reasonable. Confusing, as I saw the issue and being fixed. I try to reproduce it on Monday.

jhump:
> Confusing, as I saw the issue and being fixed. I try to reproduce it on Monday.

No worries. Let's leave this PR open. If you are able to repro, I'd like to consider other possible fixes. But if this is what it takes, I can accept it. I'm just hesitant to add features and switches to grpcui, especially if a problem can be solved easily enough elsewhere.

gfelbing:
Hello @jhump,

sorry for the late reply. 
I reproduced it locally by setting up a nginx reverse proxy with the following configuration:

```
  server {
    listen 8080;
    server_name _;
    location /grpcui {
      proxy_pass http://localhost:37577/;
    }
  }
```

Where 37577 is the port opened by grpcui pointing to the grpc hello world with reflection enabled.
Accessing grpcui directly works fine:

grpcui-direct

But accessing grpcui through the proxy fails:

grpcui-proxy

When I remove the path prefix from the proxy configuration, I the access works, of course without the prefix:

grpcui-proxy-working

jhump:
@gfelbing, you need a trailing slash for it to work that way: e.g. `localhost:8080/grpcui/`. Otherwise, a relative link to (for example) `grpc-web-form.css` resolves to localhost:8080/grpc-web-form.js, which will _not_ get routed correctly by nginx with that configuration.

Typically, the directory name, without trailing slash, would redirect to a URI _with_ the trailing slash -- so both forms end up working for end-users.

liuhuiping2013:
@gfelbing god job, continue please!

jonaskello:
Yes, I would like this too. I tried to use it to serve grpcui using the official docker image under a sub-path of my service in k8s with ingress nginx but it did not work, just got "404 page not found". I does however work if I service from "/". Would be super useful for serving under "my.service.com/grpcui".

gfelbing:
@liuhuiping2013 @jonaskello 
The work on this feature branch was basically done, the only open point was that @jhump wasn't sure if he wants this in the first place.

I rebased the branch so it could be merged again, but it's up to @jhump

jonaskello:
I finally figured out how to make it work in k8s with ingress nginx. Posting here to help anyone that struggles with the same problem. The key is to have a trailing `/` and the end. If the initial html document is loaded from `/grpcui` it will resolve relative links like `<script src="jquery-3.4.1.min.js"></script>` to `/jquery-3.4.1.min.js`. But if the initial document is loaded from `/grpcui/` (note the trailing slash) then the relative link is resolved to `/grpcui/jquery-3.4.1.min.js`.

So what I did in my ingress nginx config was to add rewrite but also redirect from `/grpcui` to  `/grpcui/` like this:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: myservice
spec:
  rules:
    - host: myservice.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: myservice
                port:
                  name: http
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: myservice-grpcui
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /$2
    nginx.ingress.kubernetes.io/configuration-snippet: |
      rewrite ^(/grpcui)$ /grpcui/ redirect;
spec:
  rules:
    - host: myservice.com
      http:
        paths:
          - path: /grpcui(/|$)(.*)
            pathType: Prefix
            backend:
              service:
                name: myservice-grpcui
                port:
                  number: 8080
```

jhump:
I've come around on this issue and am now receptive to a patch like this.

I think #103 is closer to what I want to merge (it's got a little more logic to handle some of the quirks/shortcomings of `http.StripPrefix`), so I'm closing this PR as a duplicate of that one.

PR #62, opened 2020-04-01
fix strategy for detecting if a number is an integer

### LINKED ISSUE #61: fixed32 datatype is not accepting values greater than 2147483647
as per https://developers.google.com/protocol-buffers/docs/proto3#scalar fixed32 is an unsigned int and should accept values upto 4,294,967,295. 
But the validation stops us from entering values greater than 2,147,483,647.

source:
https://github.com/fullstorydev/grpcui/blob/20f41010374fb569638f2d094776e27ffd1a0634/internal/resources/webform/webform.js#L1081

same error in java:
https://github.com/grpc/grpc-java/issues/5040

jhump:
Old approach only worked for numbers less than 2^31, which means that large uint32 and fixed32 values were rejected.

This fixes #61


internal/resources/webform/webform.js
```diff
@@ -1093,7 +1093,7 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug) {
     }
 
     function isInt(num) {
-        return (num ^ 0) === num;
+        return (num % 1) === 0;
```
dragonsinth, line 1096: maybe a comment or reference on this?  What happens for integral floats larger than 2^32 or 2^64?

jhump: done

PR #67, opened 2020-05-13
Numerous fixes and additions

jhump:
These are a variety of fixes and changes accumulated in an internal fork of this repo.

I think these about get us to a v1.0 as they includes numerous fixes, almost all around support for the JSON format of requests and the "Raw Request (JSON)" tab:
1. This was strictly requiring strings for 64-bit integers, but the JSON mapping states that numbers are acceptable. So now one can paste a JSON blob in that uses numbers, and they are handled correctly.
2. This was requiring that enum values always use the value names. The JSON mapping states that the numeric value is acceptable. So now one can paste a JSON blob in that uses enum value numbers, and they are handled correctly.
3. Field names in the JSON blob were generally wrong. It _always_ used the protobuf field name. However the JSON mapping states that the name is converted to camel-case and that the `json_name` field option overrides. So now the JSON produced by entering into the request form uses the correct JSON field names. Also, pasting a JSON blob in will be handled correctly if it uses the correct JSON field names. (If it instead uses the proto names, those are still accepted, too.)

There are also several additions:
1. To the UI and command-line tool: a new *Request Timeout* field is present on the main request form. This allows the user to control the deadline associated with the RPC that is sent.
2. To the Go package APIs: CSS, JS, and the actual HTML of the `standalone` package can now be customized. Previously, using custom CSS, JS, or HTML required falling back to the primitives in the `grpcui` package. But now they can be supplied as options to `standalone.Handler` to create a single fully-self-contained handler which serves customized content.


jhump:
@donjaime, this includes the work you did to allow customization of assets via options for `standalone.Handler`.

I omitted the method options as I have some ideas on how to do that in a different fashion. I have in mind a more uniform mechanism that provides both a nicer look and also provides that kind of information for _everything_ (e.g. an ability to also see service options, message options, field options, etc). This mechanism would also allow us to show source comments (if present in the descriptor), to provide even more info about the proto schema.

@sanjaybv, I've also upstreamed a bunch of changes to handling of the JSON tab (as enumerated in the PR description). This includes your latest fix. However, I chose to use the enum `name` as the initial value, instead of the `num`. This matches the previous behavior (before I had improved support for enums, only the name was used everywhere). It also means that the initial JSON blob, if a request has any required enum fields, will be consistent and use the name of the enum instead of its number.

@camphillips22, this includes the request timeout field you added. However, I moved to the bottom of the request form, instead of the top. WDYT?

jhump:
(@sanjaybv and @camphillips22, GitHub won't let me assign this PR to you. Not sure why not. I see you are both members of the fullstorydev org. Maybe I'd need to add you to the `fullstorydev/grpcians` group for it to allow you as assignees?)

jhump:
Green build: https://travis-ci.org/github/fullstorydev/grpcui/builds/686765388

PR #68, opened 2020-05-14
auto-open a browser pointed at the gRPCui

### LINKED ISSUE #64: Auto opening the url in default browser
Would it be possible to have an option to automatically open the url the default browser? That would be a time saver instead of copy and pasting it

jhump:
I'm still a little on the fence if it's a good idea for this to be the default behavior. I've limited it so it does _not_ try to do this by default if stdin is not a terminal. But I'm wondering if there are any other better signals as to when this is appropriate. I still need to double-check that, with the current logic, it won't try to do this when running in a container.

The `go.mod` and `go.sum` files have changed a bit more than might be expected because I ran `make ci` to re-generate it. (Last time I re-created it, I had just used `go test ./...`, so it was incomplete with regards to the various tools used in CI checks.)

Resolves #64


cmd/grpcui/grpcui.go
```diff
@@ -156,6 +163,10 @@ func (s *multiString) Set(value string) error {
 }
 
 func main() {
+	if terminal.IsTerminal(int(os.Stdin.Fd())) {
+		*openBrowser = true
+	}
+
```
dragonsinth, line 169: With a bool, there's no actual way to determine if the flag was set or was defaulted right?  So you'll just do this unconditionally?  If so, makes me wonder why make it a flag at all...

Also worth checking of stdout is a terminal instead of stdin-- stdin makes sense when deciding to prompt for confirmations.  But in this case, it's really more about what you want to do with the output (the url of the running server).

jhump: This is being set _before_ parsing flags. So we default it to true for terminals, false otherwise. Then we parse flags so command-line can override.

I was checking stdout because it seemed the more likely indicator that we are running interactive (even though we aren't prompting for input). I could see doing `grpcui api.grpc.me:443 | tee grpcui.log` and still wanting that to open in a browser.

I could also just always default to false and make people remember `-open-browser` to save themselves from clicking "new tab" and then copying+pasting the URL...

I think for me personally, I almost always want it to open the browser window. But I know there are users (like about everyone that runs it in a docker container) that use it as a sidecar to expose a UI for some other gRPC server. But I _think_ the sidecar case -- whether it's in a container or a service on a VM -- would almost always be deamonized and have stdin set to `/dev/null`.

dragonsinth: Ah. Yep probably true.

PR #72, opened 2020-05-23
add history support

### LINKED ISSUE #4: Ability to view RPC history
If the details of each RPC invocation were include in the program output/logs, it could provide great utility for someone that wants to go back to previous RPCs they've sent in the same `grpcui` session. This would be particularly slick if they were in the form of `grpcurl` invocations (see #3) and could be easily pasted into any other `grpcui` session.

It might also be nice to have this sort of history in the browser -- maybe saved to local storage so that refreshing the page doesn't forget it all. (Use of local storage would likely need to be opt-in, so other places where the form is embedded could disable that to make sure it can't interfere with the host page.)

gtg471h:
Fixes #4 

Adding local storage based history for grpcui.

This change will add a history tab that will store the last 100 requests in a rolling history.

Any history item where the service and method still exists can be loaded into the request tab to be rerun.

It supports deletion and clearing of history as needed as well.


internal/resources/webform/webform.js
```diff
@@ -2354,6 +2404,115 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug) {
...
+    }
+
+    const writeHistory = () => {
+        localStorage.setItem(storageKey, JSON.stringify(history));
+    }
+
+    const clearHistory = () => {
+        if(confirm('Are you sure you wish to delete all history? This action is permanent and cannot be undone.')) {
+            history = [];
+            writeHistory();
```
jhump, line 2428: All calls to `writeHistory` are always immediately followed by `updateHistoryUI`. Maybe they should be combined -- like rename `writeHistory` to `onHistoryChange` and have it save to local storage and then call `updateHistoryUI`?

internal/resources/webform/webform.js
```diff
@@ -2354,6 +2404,115 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug) {
...
+                </div>
+                <div>
+                    <span>MetaData:</span>
+                    <span>${JSON.stringify(item.request.metadata)}</span>
+                </div>
+            </div>`);
+            $(`#delete-${id}`).click((evt) => {
+                deleteHistoryItem(i);
+                evt.preventDefault();
+                evt.stopImmediatePropagation();
```
jhump, line 2473: What is the purpose of these? Since they are just buttons, and not submit inputs, I didn't think anything would happen on click. Maybe these are unnecessary and can be removed?

gtg471h: It prevents the onClick of the accordion header from occurring when pressing the button specifically.

jhump: Ah, I see. So if we let it propagate, the accordion component would try to collapse? How about a comment to that effect?

internal/resources/webform/webform.js
```diff
@@ -2130,46 +2147,68 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug) {
...
+                if (responseData.error) {
+                    addHistory({
+                        request: {
+                            timeout: timeoutString,
+                            metadata: metadata,
+                            data: originalData
+                        },
+                        service: service,
+                        method: method,
+                        time: new Date().toISOString(),
```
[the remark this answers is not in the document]
gtg471h: My thought was to store start time and duration if that works for you.

internal/resources/webform/webform.js
```diff
@@ -2130,46 +2147,68 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug) {
...
+                        result: 'OK'
```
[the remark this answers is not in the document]
gtg471h: Looking at this, we would have to duplicate a lot of logic to determine the status in the history UI for this to work easily by only storing the response object. How do you feel about storing the status and the response body as 2 properties of the history item?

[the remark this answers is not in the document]
jhump: I don't understand. I see almost logic here to determine the status. Looks to be basically:
```
responseData.error ? (responseData.error.name ?? 'FAILED') : 'OK'.
```
What am I missing?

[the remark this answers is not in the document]
jhump: Per convo in Slack: there is also a case when we get no reply from the server due to some other sort of error (like network failure, etc). So for each RPC we'd store something like `Either<error,ResponseData>` -- where the `error` case is for an error that prevented any sort of reply. And a `ResponseData` could still be an RPC failure -- the RPC status code and error details will be attributes contained in that object.

internal/resources/webform/webform.js
```diff
@@ -2113,11 +2142,11 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug) {
                 // to render maps a little differently and also to omit unset one-of fields
                 // (or even better, be able to render unset fields differently).
                 // But we need response schema info to do that...
-                if (data.responses instanceof Array && data.responses.length > 0) {
-                    var div = $("#grpc-response-data");
+                if (responseData.responses instanceof Array && responseData.responses.length > 0) {
+                    var div = $("#grpc-response-responseData");
```
jhump, line 2146: This stuttering looks like the result of global search+replace going too far.

internal/resources/webform/webform.js
```diff
@@ -2131,46 +2160,58 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug) {
                     }
                     // TODO(jh): button or some other mechanism to view result as raw JSON.
                 } else {
-                    $("#grpc-response-data").hide();
+                    $("#grpc-response-responseData").hide();
```
jhump, line 2163: here, too

internal/resources/webform/webform.js
```diff
@@ -2131,46 +2160,58 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug) {
...
                     } else {
                         $("#grpc-response-error-msg").hide();
                     }
                     // TODO(jh): show error details, in the same way as we
                     // show response messages above
                     $("#grpc-response-error-details").hide();
                 } else {
+                    addHistory({
+                        ...history,
+                        duration: window.performance.now() - startTime,
```
jhump, line 2189: What type is this? Will it just be a number? If so, let's name it after units (e.g. `durationMillis` or whatever), to make it clear.

internal/resources/webform/webform.js
```diff
@@ -2181,10 +2222,15 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug) {
                 t.tabs("enable", 2);
                 t.tabs("option", "active", 2);
             })
-            .fail(function(data, status) {
+            .fail(function(failureData, status) {
+                addHistory({
+                    ...history,
+                    duration: window.performance.now() - startTime,
+                    result: `failure: ${status}`
```
jhump, line 2229: Why not instead store `failureData` and `status`, instead of stringifying? What types will they be? More structure means we could make interesting improvements later w/out worrying about local storage having older representation and having to support multiple versions of data.

internal/resources/webform/webform.js
```diff
@@ -2131,46 +2160,58 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug) {
...
                         $("#grpc-response-error-msg").hide();
                     }
                     // TODO(jh): show error details, in the same way as we
                     // show response messages above
                     $("#grpc-response-error-details").hide();
                 } else {
+                    addHistory({
+                        ...history,
+                        duration: window.performance.now() - startTime,
+                        result: 'OK',
```
jhump, line 2190: I think `result` is strictly, and easily, derivable from other fields we are storing. How about we just omit it?

internal/resources/webform/webform.js
```diff
@@ -2355,6 +2401,119 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug) {
...
+        updateHistoryUI();
+    }
+
+    const onHistoryChange = () => {
+        localStorage.setItem(storageKey, JSON.stringify(history));
+        updateHistoryUI();
+    }
+
+    const clearHistory = () => {
+        if(confirm('Are you sure you wish to delete all history? This action is permanent and cannot be undone.')) {
```
jhump, line 2424: super-nit: space after `if`

internal/resources/webform/webform.js
```diff
@@ -2401,4 +2564,4 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug) {
 
     // add a single blank entry to request metadata table
     addMetadataRow();
-};
+};
```
jhump, line 2567: fix newline?

internal/resources/webform/webform.js
gtg471h: `The DOMHighResTimeStamp type is a double and is used to store a time value in milliseconds.`

I'll rename it, maybe `durationMS`?

PR #73, opened 2020-05-27
tweak Dockerfile: no need to install gcc to build modules w/ go 1.13

jhump:
With Go 1.11, the Go SDK shipped with only non-module binaries for the standard library. So building something with modules required re-building the standard library, which required having GCC and some other tools installed.

But with Go 1.13, the module versions of the standard libs are included. So we can now skip this step in setting up the Docker container for building grpcui using modules.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,6 @@
 FROM golang:1.13-alpine as builder
 MAINTAINER FullStory Engineering
 
-# currently, a module build requires gcc (so Go tool can build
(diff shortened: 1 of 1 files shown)


jhump:
CI won't be green until after I can land #74...

Dockerfile
```diff
@@ -1,10 +1,6 @@
 FROM golang:1.13-alpine as builder
```
jameremo, line 1: So this PR is a follow-up to #70.

PR #74, opened 2020-05-27
fix CI again... use Go 1.13 to run vet tools

jhump:
I'm not sure why this suddenly started breaking with Go 1.11.

Maybe related to the fact that, for some reason I cannot determine, the `go` tool keeps wanting to upgrade the `predeclared` tool?

So this also updates the `go.mod` file to reflect that upgrade and also includes associated `go.sum` changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,7 @@ matrix:
     - go: 1.11.x
       env: GO111MODULE=off
     - go: 1.11.x
-      env:
-      - GO111MODULE=on
-      - VET=1
+      env: GO111MODULE=on
     - go: 1.12.x
       env: GO111MODULE=off
     - go: 1.12.x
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -7,25 +7,19 @@ matrix:
...
+      env: GO111MODULE=on
     - go: 1.12.x
       env: GO111MODULE=off
     - go: 1.12.x
       env: GO111MODULE=on
     - go: 1.13.x
       env:
       - GO111MODULE=on
-      - GOPROXY=direct
+      - VET=1
```
jameremo, line 18: Why vet 1.13, why not the latest version of go?

.travis.yml
```diff
@@ -7,25 +7,19 @@ matrix:
...
     - go: 1.14.x
-      env:
-      - GO111MODULE=on
-      - GOPROXY=direct
```
jameremo, line 24: What's different that we now want to use the default proxy here? It's been around since 1.13 came out, right?

jhump: Sorry for lack of context. Some of this cleanup was just making things consistent with other repos. I've removed the `GOPROXY=direct` so that CI will use the Go proxy servers, instead of always trying to directly source control.

This env var was added once upon a time when CI was red on "tip" version (maybe some instability?). Now that things seem stable on all recent versions (basically 1.13 and up), I figured using the default Go proxy was reasonable.

.travis.yml
jhump: Sorry, I forgot to answer this one before merging. I've been sticking with versions that we happen to use internally, to make it easier to repro on our work laptops.

PR #76, opened 2020-06-22
improve some sharp edges of request timeout validation

jhump:
This improves the UX for the request timeout text area.

When you use the `type="number"` attribute on an input field, getting the value suddenly coerces the contents to a number. That means there is no easy way to distinguish between the input being empty vs. the input having a zero in it.

Also, if you tabbed into the input and tried to tab out without typing a value, you'd get an error alert. Due to the quirk described above, it would see a zero and show an error that the timeout must be greater than zero.

So I removed that attribute and added some checks in the code for the case of the input being empty (to treat that as `undefined` -- e.g. not specified, so use default/none). I also noticed that when there was an error, the validation was clearing out the input. So if I typed "3" (valid), and then clicked back in and changed it to "foo", it would clear the input. That seems unfriendly, so I've fixed it to instead _restore_ the prior value. So in the scenario just described, after typing "foo" and getting an error about that not being a number, it would restore the contents to "3".


internal/resources/webform/webform.js
```diff
@@ -2356,21 +2357,23 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug) {
...
-            var num = + $(inp).val();
-            if (num === undefined) {
+            var val = $(inp).val();
+            if (val === "" || val === undefined) {
                 return;
             }
+            var num = Number(val);
             if (Number.isNaN(num)) {
-                $(inp).val(undefined);
+                $(inp).val(origValue);
```
patrick-fs, line 2368: I guess one pitfall here is that the user could see the validation error message "numeric value required" while _also_ seeing the original numeric value in the input? The end result being "I see a thing telling me a numeric value is required, but I also see a numeric value in the input, wha?"

Am I reading this right?

jhump: Interesting point. In practice, at least in Chrome, this isn't an issue. I think it won't actually re-render this DOM change until after the JS thread quits (and since `alert` is blocking/modal, it doesn't actually update the page until after the modal is dismissed). However, even if that were to change, I think a better fix is to include the invalid text in the error message. That way, even if they see it change back, it will show them the invalid string they had entered. Does that seem like a reasonable solution?

jhump: Hmmm, actually, it looks like there is some intent that the bad value be left. It actually applies a CSS attribute, to highlight the input as invalid (using same code used to validate inputs in the request form and to highlight invalid fields). So I can change it to leave the value and then fix the CSS so it's properly highlights, and that will probably make for the best experience.

PR #80, opened 2020-07-10
fix issue when repeated/map field is nested inside a map field value

jhump:
This addresses an issue when trying to add elements to a repeated or map field that is _nested inside_ the value of another map field.

The enclosing map field was overriding the `onAdd` event, but incorrectly: it assumed that it would only be called when entries were _directly_ added to the map. But it also gets called when elements are added to descendants (like a nested repeated field, for example).

So it failed to propagate the event, which in turn failed to keep the internal request model in sync with the form.

This fixes it. Now, the map's `onAdd` special-cases when the event is adding an entry directly to the map. And otherwise propagates the event in the same fashion as done with `onChange` events.


jhump:
@mattfs, this is the patch you reviewed earlier this week, upstreamed to the source repo.

Issue #85, opened 2020-08-27
Specify headers to use regardless of reflection


rosshadden:
As the docs say, `-reflect-header`s are only used for reflection requests:
> These headers will only be used during reflection requests

Is there a way to pass headers that get used regardless?

jhump:
@rosshadden, `-H` sets headers used for all requests.

rosshadden:
@jhump Is that an upcoming change? The version I have installed (latest from `go get`) doesn't have or support that option, and I can't find it in the source code in `master` either.

jhump:
Oh weird. How did we lose that... I thought it was there all along. So it's an accidental omission. It needs to be added (back?)

jhump:
Oh wait... Sorry, I was looking at `grpcurl -h` when I saw it.

It was in fact intentional that it not be in `grpcui`: because the user specifies the metadata in the UI.

I guess if we did have a `-H` option, it would just be used to populate the default values in the UI. I certainly wouldn't want there to be anything passed to the server that the user couldn't see or control. Maybe it should be called `-default-headers` instead.

The `-reflect-headers` are still needed because they are issued transparently -- the user does not invoke them through the UI but rather they are executed _before_ the UI can be shown to populate the drop-downs and schema info.

rosshadden:
I like the `-default-headers` idea, and understand that you want them to be seen and controllable in the UI :+1:.

jhump:
Addressed in #88.

PR #88, opened 2020-09-21
add ability to set default metadata; fix loading of metadata and stream requests from history

### LINKED ISSUE #84: Loading metadata from history
Loading request from history works well, but not for metadata.
Is it known issue?

jhump:
This fixes a recently introduced issue where the attempt to load metadata from history fails. (It also would fail to load request data for client-streaming RPCs.) The issue is that it was trying to deep-copy a value, but it assumed the value to copy was always an object, when it could instead be an array.

This also introduces `-default-headers` and `-debug-client` flags. The former lets you set default headers that show up in the metadata form whenever the grpcui is refreshed (they are editable/deletable by the user). The latter enables some console logging in the JS client (previously enabled only via environment variable).

This fixes #84 and #85/#87.


jhump:
@gtg471h, mind taking a look?

gtg471h (approved):
LGTM

focusaurus:
@jhump Would you mind publishing a release for this?

PR #91, opened 2020-10-08
css/ux tweaks for history tab

jhump:
I'm about to make a release and wanted to do some cleanup of the history tab, to improve the look a little and also make sure the style and feel was more like the rest of the UI.

There were a few default jqueryui styles in there that were jarringly different IMO, and one that was causing an annoying 404 error because it was trying to download an icon that grpcui does not serve. I also tried to make things line up a little better, more like a table so the column runs are straight instead of wavy.

Below are a few screenshots that show basically all possible styles that have been changed by this PR.

----

### Default view when opening tab (nothing selected):
Screen Shot 2020-10-08 at 4 15 36 PM

----

###  Expanding an item that had no metadata set:
Screen Shot 2020-10-08 at 4 15 50 PM

----

###  Expanding an item that did have custom metadata:
Screen Shot 2020-10-08 at 4 16 04 PM


jhump:
For reference, here's a screenshot using `grpcui` from master with the same data as the screenshots above:

----

### Defaults to first element expanded:
Screen Shot 2020-10-08 at 4 03 35 PM

----

### After collapsing first element:
(The "_Service or method no longer available_" in this screenshot isn't shown in above screenshots because it is now a tooltip if you hover over struckthrough method name.)
Screen Shot 2020-10-08 at 4 03 52 PM

gtg471h (approved):
LGTM

PR #95, opened 2020-10-22
Basic support for loading/saving history

cosmin-mogos:
A simple approach to inject history into grpcui to act as a collection


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.13
 require (
 	github.com/fullstorydev/grpcurl v1.7.0
 	github.com/go-bindata/go-bindata v0.0.0-20191126083508-8639be0519b3 // indirect
-	github.com/golang/protobuf v1.3.5
+	github.com/golang/protobuf v1.4.1
```
jhump, line 8: It doesn't look like you are actually doing anything that requires any updates to `go.mod`.

Due to compatibility issues between 1.3.5 and 1.4.1, I'd rather not update this repo just yet. (You can, of course, use v1.4.1 in our own module that uses grpcui; but I don't want to foist the new version on _every_ user yet.)

go.mod
```diff
@@ -15,5 +15,6 @@ require (
 	golang.org/x/net v0.0.0-20200324143707-d3edc9973b7e
 	google.golang.org/api v0.29.0 // indirect
 	google.golang.org/grpc v1.30.0
+	google.golang.org/protobuf v1.25.0
```
jhump, line 18: Ditto. Let's just revert the changes to both `go.mod` and `go.sum`.

internal/resources/webform/webform.js
```diff
@@ -2424,6 +2425,19 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug, headers)
...
+
+        $.ajax({
+            url: 's/history.json',
+            type: 'GET',
+            success: function(data){
+                preloadedHistory = data;
+                updateHistoryUI();
+            },
+            error: function(data) {
+                // nothing to do probably no history saved server side
```
jhump, line 2437: I think we should actually have a stub in the server that can return an empty JSON document, even if the user has not registered a history file. In fact, to make this less subtle and more discoverable in the Go API, I think it best to add first class support for it to the `standalone` package, in the form of a new `HandlerOption`. That way, you also don't need the "/s/" prefix.

internal/resources/webform/webform.js
```diff
@@ -2435,48 +2449,67 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug, headers)
     const clearHistory = () => {
         if (confirm('Are you sure you wish to delete all history? This action is permanent and cannot be undone.')) {
             history = [];
+            preloadedHistory = [];
```
jhump, line 2452: I don't think this is necessary. Instead, when you load the server history, if it is not empty, just merge it into `history`. I think that would let you undo some of the changes/refactoring in `updateHistoryUI()` below and also return to the much simpler handling of `history` with a single index. That would allow this PR to become smaller and much simpler.

internal/resources/webform/webform.js
```diff
@@ -2435,48 +2449,67 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug, headers)
...
+        document.body.appendChild(element);
+
+        element.click();
+
+        document.body.removeChild(element);
+    }
+
+    const saveHistory = () => {
+        let toSave = history
+        if (preloadedHistory.length > 0 && confirm('Include pre-loaded (gray) items?')) {
```
jhump, line 2472: I don't think this is necessary: 1) the JSON is not that hard to edit, to remove any unwanted items and 2) a user could also manually curate by first clearing history, then issuing RPCs, and then saving.

This seems to be the only reason that you have `preloadHistory` as a separate var from `history`. And I think the change is easier to read and follow without it.

jhump (commented):
I like the approach! I have several comments.

TLDR:
* I don't think the distinction between "history" and "preloadedHistory" is that useful or necessary, and removing it would simplify this change a lot.
* Let's add proper API for setting the history via Go code. And we can do so in a way that will prevent the annoying 404 error from popping up in the console for every user of the command-line tool.

internal/resources/webform/webform.js
cosmin-mogos: Merging was my initial approach, but it ended up in a positive feedback loop that increases history with every page reload.

This is a pretty rough draft I used to get an idea of what changes would be needed. My intent is to mirror our process for HTTP: each service has a postman collection saved in the source code so new devs or clients of our service can quickly get up to speed with the api.

It would be beneficial to separate history from "official" examples, the expectation is that examples are known to work (or fail in a documented fashion) while history is simply there to give access to old calls. In the current draft the separation between examples and history is highlighted through the gray background in the image below.

Screenshot 2020-10-23 at 09 56 26

A more elaborate solution would be to add explicit examples support, in the UI this would be a list on the left side of the request tab. Clicking on an example the request form is populated with the data from the example, history will work unchanged as before. From the back end code, we'll add a dedicated `/examples` endpoint that can be populated by an option, e.g. `standalone.WithExamples(exablesBlob)`, if the option is not provided `/examples` will return an empty array which will lead the UI to not create the examples list from the examples tab.

Data schema will use the same format as history plus a description part where we can add a short name and detailed description.

internal/resources/webform/webform.js
cosmin-mogos: I've added `standalone.WithExamples` for this role.

internal/resources/webform/webform.js
jhump: Ok. I didn't think through it enough to foresee the ever-enlarging-history problem :)

internal/resources/webform/webform.js
jhump: I guess if we need to split out the pre-loaded history for other reasons, this can stay.

But I like the idea of calling them "example" instead of "pre-loaded items".

Also, instead of just gray background, would it make more sense for this to be very explicit and be in a separate table with a heading that read "Examples"?

standalone/opts.go
```diff
@@ -80,6 +80,13 @@ func ServeAsset(filename string, contents []byte) HandlerOption {
 	})
 }
 
+// WithExamples will add examples to the UI. The examples blob is expected to be valid json.
+func WithExamples(examples []byte) HandlerOption {
```
jhump, line 84: I don't think `[]byte` is sufficiently type-safe. Requiring a blob is both unsafe but also makes the API much harder to use for a casual reader. Simply "valid json" isn't enough either.

It should be pretty easy to model the actual type here with a Go struct w/ "json" struct tags. For your case, of just loading it from a JSON file, you can unmarshal it into the Go type, which provides a nice start-up check that the file is valid.

internal/resources/webform/webform.js
jhump: For now, how about renaming `preloadedHistory` -> `examples`? And I think it may also make sense to move them to a separate HTML table with their own heading, to make it more clear what they are (instead of relying on just a gray background).

cosmin-mogos: I've updated the UI to handle examples in a cleaner way. If `standalone.WithExamples` is used the UI will look like the image. If no examples are present the examples column is hidden. This approach is similar to postman and other similar tools so devs should understand it easily.
Screenshot 2020-10-23 at 17 44 47

cosmin-mogos: I also plan to add some documentation once the flow is in good state.

internal/resources/webform/webform.js
cosmin-mogos: Completely removed preloadedHistory

standalone/opts.go
cosmin-mogos: Will work on this next. I'll add a `standalone.ParseExamplesFile(examplesFilePath)`, and `standalone.ParseExamples(examplesBlob)` that output a go json model.

cosmin-mogos:
The build for 1.13 is failing for unrelated issues, how should we handle it? https://travis-ci.com/github/fullstorydev/grpcui/jobs/406363129

```
staticcheck ./...
handlers.go:16:2: package github.com/golang/protobuf/jsonpb is deprecated: Use the "google.golang.org/protobuf/encoding/protojson" package instead.  (SA1019)
handlers.go:17:2: package github.com/golang/protobuf/proto is deprecated: Use the "google.golang.org/protobuf/proto" package instead.  (SA1019)
testing/cmd/testsvr/testsvr.go:13:2: package github.com/golang/protobuf/proto is deprecated: Use the "google.golang.org/protobuf/proto" package instead.  (SA1019)
Makefile:52: recipe for target 'staticcheck' failed
```

cosmin-mogos:
The PR should be in a cleaner state, please let me know if other changes are needed.

internal/resources/webform/webform.js
```diff
@@ -2528,26 +2582,34 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug, headers)
         });
     }
 
-    const loadHistoryItem = (index) => {
+    const loadRequest = (request) => {
         const t = $("#grpc-request-response");
         t.tabs("option", "active", 0);
-        const item = history[index];
-        $("#grpc-request-timeout input").val(item.timeout);
```
jhump, line 2535: Oh, looks like you fixed a bug here too!

standalone/examples.go
```diff
@@ -0,0 +1,46 @@
...
+			Name  string `json:"name"`
+			Value string `json:"value"`
+		} `json:"metadata"`
+		Data map[string]interface{} `json:"data"`
+	} `json:"request"`
+	Service string `json:"service"`
+	Method  string `json:"method"`
+}
+
+// Examples list of gRPC request examples
```
jhump, line 23: I don't think we need any of the stuff below this point in the file. This type and the helpers below needlessly broaden the surface area of the API without providing much value.

standalone/opts.go
```diff
@@ -80,6 +80,13 @@ func ServeAsset(filename string, contents []byte) HandlerOption {
 	})
 }
 
+// WithExamples will add examples to the UI. The examples blob is expected to be valid json.
+func WithExamples(examples *Examples) HandlerOption {
```
jhump, line 84: Let's just use `...Example` here. No need for the extra wrapper type.

standalone/standalone.go
```diff
@@ -118,6 +122,24 @@ func Handler(ch grpcdynamic.Channel, target string, methods []*desc.MethodDescri
...
 }
 
+func registerExamplesHandler(mux *http.ServeMux, uiOpts *handlerOptions) {
+	var examplesBlob []byte
+	if uiOpts.examples != nil && len(*uiOpts.examples) != 0 {
+		marshaled, err := json.Marshal(uiOpts.examples)
+		if err == nil {
+			examplesBlob = marshaled
+		} else {
+			log.Printf("Failed to marshal examples: %v\n", err)
```
jhump, line 132: This is not appropriate. You could instead save the error and use that to write an error code in response to `/examples` requests.

standalone/standalone.go
```diff
@@ -118,6 +122,24 @@ func Handler(ch grpcdynamic.Channel, target string, methods []*desc.MethodDescri
 	})
 }
 
+func registerExamplesHandler(mux *http.ServeMux, uiOpts *handlerOptions) {
+	var examplesBlob []byte
```
jhump, line 126: I think this should default to `"[]"` when there are no examples. So if the `uiOpts.examples` type were `[]Example`, you could just unconditionally marshal.

jhump (commented):
Hey, this looks very cool. I'll have to pull it down and play with it to get a sense for the LnF and the new CSS and DOM structure. I left a couple of remarks though for shrinking/simplifying the Go API in `standalone`.

internal/resources/webform/webform.js
cosmin-mogos: :)

standalone/examples.go
```diff
@@ -0,0 +1,16 @@
+package standalone
+
+// Example model of an example gRPC request
+type Example struct {
+	Name    string `json:"name"`
+	Request struct {
+		Timeout  string `json:"timeout"`
```
jhump, line 7: Instead of making this whole structure stringly-typed, we should provide more typing, to prevent bad data from being supplied.

For this we want a `time.Duration`. But it needs to be wrapped in a type with a custom `MarshalJSON` method to marshal it to a JSON string in the format `%fs` (e.g. floating point duration with a trailing "s" indicating a unit of seconds).

standalone/examples.go
```diff
@@ -0,0 +1,16 @@
...
+		Metadata []struct {
+			Name  string `json:"name"`
+			Value string `json:"value"`
+		} `json:"metadata"`
+		Data map[string]interface{} `json:"data"`
```
jhump, line 12: This doesn't look right. Where did this come from?

This can actually be a primitive, like if the RPC request type is `google.protobuf.StringValue` or `google.protobuf.FloatValue`.

And it can be an array, if the method in question accepts a stream from the client (each element in the array is a message in the stream).

So I think we instead need `interface{}`. _However_, it is highly likely that someone constructing the example programmatically might want to construct an actual proto message and put it here. And proto messages do not correctly support marshaling via `encoding/json`. So I think this whole `Request` field needs to pulled out to a named type that has a custom `MarshalJSON` function, which will use `github.com/protobuf/jsonpb` to marshal the value(s) if it is a `proto.Message` or is a slice or array with `proto.Message` elements.

standalone/standalone.go
```diff
@@ -118,6 +121,22 @@ func Handler(ch grpcdynamic.Channel, target string, methods []*desc.MethodDescri
...
+func registerExamplesHandler(mux *http.ServeMux, uiOpts *handlerOptions) {
+	examplesBlob, err := json.Marshal(uiOpts.examples)
+
+	mux.HandleFunc("/examples", func(w http.ResponseWriter, r *http.Request) {
+		if err == nil {
+			w.Header().Add("Content-Type", "application/json")
+			w.WriteHeader(200)
+			w.Write(examplesBlob)
+		} else {
+			w.Header().Add("Content-Type", "text/plain")
```
jhump, line 133: I know I suggested this. So thank you for making the change. But I'm now wondering if we should actually push the marshalling to JSON all the way up to the call to `WithExamples` (when building the set of options) and just `panic` at that point if we can't marshal the examples. It's a little subtle and possibly bad user experience that we provide no feedback about bad examples provided until a browser tries to fetch them.

What do you think?

jhump (commented):
Hey, sorry to keep dropping comments. I realized an issue with the way the `Example` type is defined, that makes it harder to use programmatically. Sorry if this is starting to feel like scope-creep.

standalone/examples.go
cosmin-mogos: Makes sense to me, if we complicate encoding rules, I would add back the decoding helpers as well to make it easy for users to handle example files.

standalone/standalone.go
cosmin-mogos: good point, encoding errors should be a rare exception, if they happen then fail fast.

standalone/examples.go
cosmin-mogos: I'll experiment with this.

cosmin-mogos:
> Hey, sorry to keep dropping comments. I realized an issue with the way the `Example` type is defined, that makes it harder to use programmatically. Sorry if this is starting to feel like scope-creep.

Two things need to be done
1) update examples handling
2) fix the build
After these changes is there something else that needs to be adjusted?

How can travis be fixed? For go 1.13 the failure is 
```
staticcheck ./...
handlers.go:16:2: package github.com/golang/protobuf/jsonpb is deprecated: Use the "google.golang.org/protobuf/encoding/protojson" package instead.  (SA1019)
handlers.go:17:2: package github.com/golang/protobuf/proto is deprecated: Use the "google.golang.org/protobuf/proto" package instead.  (SA1019)
testing/cmd/testsvr/testsvr.go:13:2: package github.com/golang/protobuf/proto is deprecated: Use the "google.golang.org/protobuf/proto" package instead.  (SA1019)
Makefile:52: recipe for target 'staticcheck' failed
```

standalone/examples.go
jhump: I wasn't suggesting we complicate anything -- users should still be able to use `json.Marshal` to read/write files. We put the complexity behind the `JSONMarshaler` interface abstraction. Also, I'm just not a fan of helpers that just call `io.ReadFile` and then call something else. That's so trivial for caller to do that it doesn't file like it carries its weight in a library.

standalone/examples.go
jhump: But even if you save the file and always unmarshal from JSON, if your RPC schema has a streaming request or one of the wrapper types as a request, this type won't work and instead needs to be `interface{}`.

I'm fine just having a TODO for now and leaving out the `jsonpb` stuff I suggested. But, to make sure I don't need to make any backwards incompatible changes to address that TODO, I think you need to hoist the `Request` field out to its own named type, instead of an anonymous struct type.

Personally, I can see creating examples programmatically being a very reasonable thing for someone to try upon seeing this API; and they'd be disappointed that it didn't work.

jhump:
> After these changes is there something else that needs to be adjusted?

I don't think so.

> How can travis be fixed? For go 1.13 the failure is

You're not using the right version of `protoc-gen-go` when you run `make generate`. You have changes to generated proto code that pull in a newer version of the protobuf runtime. (I have intentionally not upgraded this repo yet because the move to 1.4.0 isn't small -- tons of deprecation warnings to fix are just a part of it.)

So you could just revert the change to `testing/cmd/testsvr/test.pb.go` and that should fix it.

It only happens for Go 1.13 because that is the version for which our CI setup _also_ runs a bunch of other vet checks (which is what is catching this).

standalone/examples.go
cosmin-mogos: Sorry if this sounded defensive, it was not the intent :). For a library the requirements are tighter, it makes sense to have better tools around examples. 

Please have a look at the current draft, I plan to add more tests and handle gRPC payload.

cosmin-mogos:
> You're not using the right version of `protoc-gen-go` when you run `make generate`

Which version should I use? I've added the testmodels.proto used in examples_tests.go and will need to generate the appropriate code.

jhump:
> Which version should I use?

The one already indicated in the `go.mod` file. So this should install the right version:

```
go install github.com/golang/protobuf/protoc-gen-go
```

(Relevant: #98)

cosmin-mogos:
Updated the examples handling please review and let me know what the next steps are.

cosmin-mogos:
The PR should be in it's final stages please have a look.

standalone/examples.go
```diff
@@ -0,0 +1,135 @@
...
+	jsm := jsonpb.Marshaler{EmitDefaults: true, OrigName: true}
+	var b bytes.Buffer
+	if err := jsm.Marshal(&b, msg); err == nil {
+		return nil, err
+	}
+	return b.Bytes(), nil
+}
+
+func (request *Request) UnmarshalJSON(b []byte) (returnedErr error) {
+	var raw map[string]interface{}
```
jhump, line 98: Instead of this very manual unmarshal, this could be improved in a couple of ways:

1. As mentioned above, if you change `Timeout time.Duration` to `TimeoutSeconds float32`, then you don't even need a custom unmarshal function at all. (I know I suggested `time.Duration`; I usually prefer type-safety. But I offer undoing that suggestion since I think the resulting code would be simpler).
2. OR, if you like type safety (who doesn't, right?), we could leave it `time.Duration`, but then use the technique described above for `Marshal`, where you unmarshal into a struct that has the exact right JSON format, and then convert from there. That way, you are only converting/adapting the field(s) that need it instead of having to write a map unmarshaller like this.

standalone/examples_test.go
```diff
@@ -0,0 +1,200 @@
+package standalone
+
+import (
+	"encoding/json"
+	"github.com/fullstorydev/grpcui/testing/testdata"
```
jhump, line 5: ditto about `goimports`

standalone/standalone.go
```diff
@@ -118,6 +120,18 @@ func Handler(ch grpcdynamic.Channel, target string, methods []*desc.MethodDescri
 	})
 }
 
+func registerExamplesHandler(mux *http.ServeMux, uiOpts *handlerOptions) {
+	mux.HandleFunc("/examples", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Add("Content-Type", "application/json")
+		w.WriteHeader(200)
+		if len(uiOpts.examples) > 0 {
```
jhump, line 127: nit: you could resolve this outside the call to `HandleFunc`, so the actual handler just blindly writes bytes.

testing/testdata/testmodels2.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto2";
```
jhump, line 1: Let's move all of these to `internal/testdata`, so the resulting Go code is not part of the exported API of this repo.

standalone/examples_test.go
jhump: I'm confused how this compiles. I don't see any references to `testdata` in here.

testing/testdata/testmodels2.proto
```diff
@@ -0,0 +1,12 @@
...
+
+package testmodels;
+
+option go_package = ".;testmodels";
```
jhump, line 5: Eek! Please don't generate Go files where the package is needlessly different than the last element of package path.
```suggestion
option go_package = ".;testdata";
```

testing/testdata/testmodels2.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto2";
+
+package testmodels;
```
jhump, line 3: For consistency, let's make this `testdata`, too.

standalone/examples_test.go
```diff
@@ -0,0 +1,200 @@
...
+				t.Fatalf("aInt32 == %v != 107", data["aInt32"])
+			}
+			if data["aString"] != "string" {
+				t.Fatalf("aString == %v != \"string\"", data["aString"])
+			}
+		})
+	}
+}
+
+func strPtr(s string) *string {
```
jhump, line 165: You don't need either of these functions. Use `proto.String` and `proto.Int32`.

standalone/examples_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			marshal, err := json.Marshal(test.request)
+			if err != nil {
+				t.Fatalf("marshal failed: %v", err)
+			}
+			t.Logf("marshaled: %q", string(marshal))
+
+			var raw interface{}
```
jhump, line 148: If you unmarshal into a custom struct, you don't have to do the gross checks below but can instead just use `proto.Equal` to compare the unmarshaled proto to the original.

standalone/examples_test.go
```diff
@@ -0,0 +1,200 @@
...
+				Data: testmodels.TestMessage3{
+					AInt32:  107,
+					AString: "string",
+				},
+			},
+		},
+		{
+			name: "proto2 data",
+			request: Request{
+				Data: testmodels.TestMessage2{
```
jhump, line 132: I think we don't even need any custom test protos here -- you could instead test using well-known protos, like `FieldDescriptorProto` and `Struct`.

testing/testdata/testmodels2.proto
jhump: This is confusing that you have two test files with "2" and "3" suffix. What happened to number one? They should both be combined into `testdata.proto`. OR you could just use well-known types (that are part of the protobuf runtime library) and not even bother with custom messages here.

standalone/standalone.go
```diff
@@ -99,6 +99,8 @@ func Handler(ch grpcdynamic.Channel, target string, methods []*desc.MethodDescri
 	rpcMetadataHandler := grpcui.RPCMetadataHandler(methods, files)
 	mux.Handle("/metadata", rpcMetadataHandler)
 
+	registerExamplesHandler(&mux, uiOpts)
```
jhump, line 102: While I understand this makes the code more modular and more concise, it's also inconsistent with the rest of this method -- you've moved the handler out to a separate function even though it's no more complicated/no larger than any of the other handlers inlined into this function. Consider inlining.

jhump (commented):
I have some suggestions on improving the JSON marshaling/unmarshaling. Also, I have some requests for cleaning up the test protos you added. Otherwise, I do like it.

jhump:
Sorry it took me so long to review your latest updates. It's been a very busy week.

cosmin-mogos:
> I have some suggestions on improving the JSON marshaling/unmarshaling. Also, I have some requests for cleaning up the test protos you added. Otherwise, I do like it.

This will be the last round of changes from my side, if we don't reach consensus I'll have to keep using the fork, which is working fine for us in production.

standalone/examples.go
```diff
@@ -0,0 +1,135 @@
...
+
+// MetadataPair (name, value) pair
+type MetadataPair struct {
+	Name  string `json:"name"`
+	Value string `json:"value"`
+}
+
+// Request gRPC request
+type Request struct {
+	Timeout  time.Duration
```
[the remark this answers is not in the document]
cosmin-mogos: String was working in the previous version. Since we're using custom marshaling/unmarshaling and the code is already implemented and tested I'd prefer to keep it as is.

standalone/examples_test.go
cosmin-mogos: Can you please suggest an example?

internal/resources/webform/webform.js
```diff
@@ -2414,11 +2414,48 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug, headers)
...
+
+    const updateExamplesUI = () => {
+        // only populate the example list if we have some
+        if (examples.length == 0) {
+            return
+        }
+
+        let examplesList = $("#grpc-request-examples")
+        examples.forEach(example => {
+            let exampleItem = $(`<li class="grpc-request-example">${example.name}</li>`)
```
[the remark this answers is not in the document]
cosmin-mogos: I'm also no expert in js :). All I can say is that the code executed in my version of chrome and firefox without throwing exceptions. I'll add the semicolon to make it more explicit.

testing/testdata/testmodels2.proto
cosmin-mogos: Do we keep the custom data which allows for differentiation between proto2 and 3 or do you prefer to use predefined types as suggested in https://github.com/fullstorydev/grpcui/pull/95#discussion_r518776961.

I think there's some benefit to having tests for proto 2 and 3, if 2 is dropped at some point in the future failing tests can help signal it.

standalone/examples_test.go
cosmin-mogos: Not sure I understand how we can decode a `proto.Message` in `ExampleRequest.Data` as implemented now, `Data` will contain whatever json.Decode returned. We'd need to add an extra step to convert from `interface{}` to `proto.Message`. Can you please give an example?

jhump: You currently aren't using `ExampleRequest`, so would continue to do that. So instead of unmarshaling into `interface{}`, you'd unmarshal into a 
```go
// this happens to exactly match the struct in the main `github.com/fullstorydev/grpcui` package
// that is used to parse request that is sent from the browser...
type request struct {
    TimeoutSeconds float32 `json:"timeout_seconds"`
    Metadata []ExampleMetadataEntry `json:"metadata"`
    Data json.RawMessage `json:"data"`
}
// You can then use jsonpb to unmarshal req.Data into whatever protobuf message and
// then use proto.Equal to verify that the unmarshaled proto matches the one that was
// originally marshaled.
```

testing/testdata/testmodels2.proto
jhump: >  if 2 is dropped at some point in the future

That can't realistically happen since `descriptor.proto`, which is at the very heart of all dynamic protobuf and RPC handling and even at the heart of protoc itself (as that is the format used by all protoc plugins), is proto2 (and must be, since it makes heavy use of extensions for custom options).

standalone/examples_test.go
jhump: I did: `FileDescriptorProto` (in `github.com/golang/protobuf/protoc-gen-go/descriptor`) and `Struct` (in `github.com/golang/protobuf/ptypes/struct`):
```go
// maybe something like so
Data: &descriptor.FileDescriptorProto{
    Name: proto.String("foo.proto"),
    MessageType: []*descriptor.DescriptorProto{
        {
            Name: proto.String("Foo"),
        },
    },
}
```

Also, looking more closely at this, it's not clear why this works since `testmodels.TestMessage2` is not a `proto.Message`, the _pointer_ to it (`*testmodels.TestMessage2`) is. I think this ends up using `encoding/json` -- which works if you are just trying to round-trip the value to JSON and back, but does not work to send a message to the browser (`encoding/json` may not encode it in a way that the browser client accepts).

Anyhow, use of a well-known type just keeps this repo a little smaller, leaning on existing things instead of creating new ones... Not a blocking comment, as long as it leads to no extra exported API.

cosmin-mogos: Works for me I'll delete the custom types and use `descriptor.DescriptorProto`

testing/testdata/testmodels2.proto
cosmin-mogos: Removed the custom messages completely

standalone/examples_test.go
cosmin-mogos: updated tests

standalone/examples_test.go
cosmin-mogos: Done

standalone/examples.go
```diff
@@ -0,0 +1,135 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return marshalData, nil
+}
+
+func toJSON(msg proto.Message) ([]byte, error) {
+	jsm := jsonpb.Marshaler{EmitDefaults: true, OrigName: true}
```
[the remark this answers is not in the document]
cosmin-mogos: removed

standalone/examples.go
```diff
@@ -0,0 +1,135 @@
...
+type Request struct {
+	Timeout  time.Duration
+	Metadata []MetadataPair
+	Data     interface{}
+}
+
+func (request Request) MarshalJSON() ([]byte, error) {
+	var encFields []string
+
+	if request.Timeout.Seconds() != 0 {
```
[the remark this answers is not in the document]
cosmin-mogos: makes sense, updated

standalone/standalone.go
cosmin-mogos: will work either way, and it's not really a performance issu

cosmin-mogos:
Good enough or abandon?

jhump:
> Good enough or abandon?

I apologize it's been so long without anyfeedback. Last week I attended GopherCon (virtually) and this week, we're doing a lot of Black Friday/Cyber Monday prep at work, so I haven't had time to look.

I'll try to look today or tomorrow. If for any reason I have more input, I could always pull down this branch and open a new pull request from my own branch if you are unable to address any further feedback.

cosmin-mogos:
Let's agree on the steps forward. I want to take this PR to the finishing line as I think it's a useful feature to have, to achieve this what still needs to be done?

From my point of view I can contribute bug fixes for any functionality broken by this PR.

jhump:
Hey, @cosmin-mogos, thank you for your patience. I apologize it's been so long since I've taken another look. I do intend to get this merged. And if you weren't available to make any additional requested changes, I could always make them.

However, I've been on vacation for the past week, and then busy at work with Black Friday/Cyber-Monday prep, and then before that was GopherCon (virtual this year). So I'm afraid these have led me to neglecting this PR. I will get back to it this week.

PR #100, opened 2020-11-11
Response: Render response as a JSON blob in a textarea

theMPatel:
By popular demand and also because I just wanted it


internal/resources/webform/webform.js
```diff
@@ -2156,22 +2156,7 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug, headers)
...
-                        div.append(container);
-                        var val = resp[i];
-                        if (val.isError) {
-                            container.html('<div class="error">Server error processing message #' + (i+1) + '</div>');
-                        } else {
-                            populateResultContainer(container, 0, val.message);
-                        }
-                    }
-                    // TODO(jh): button or some other mechanism to view result as raw JSON.
+                    $("#grpc-response-textarea").val(JSON.stringify(responseData.responses, null, 2));
```
theMPatel, line 2159: @jhump kind of juggled with this one, i would be ideal if we didn't have to show the whole `[ { message : { .. }, isError: true/false } ]` structure, but couldn't come up with a good way to handle when the request for a particular message errored. This just dumps the whole payload

internal/resources/webform/webform-sample.css
```diff
@@ -411,6 +411,11 @@ table#grpc-request-metadata-form {
     padding: 0;
 }
 
+textarea#grpc-response-textarea {
+    font-family: Monoca, monospace;
```
jhump, line 415: Let's please be consistent and just use the same font/formatting as other text areas.

internal/resources/webform/webform.js
jhump: That will be very misleading -- if someone wanted to serialize the output back into a response proto, they'd have to know how to mangle this output/strip the envelope.

Also, we don't really want to show a JSON array if the actual response data is not an array (either via use of `google.protobuf.ListValue` as the response type, or using a server stream).

One thing we could do is to create multiple divs, one for each response message. And that div contains either some sort of error signal (like it did before) _or_ it contains a textarea that shows the JSON for that particular response message. For most RPCs this will be quite natural and an ideal way to represent this output. For server streams, it could admittedly be a little awkward, since you can't just copy the entire response stream. But that is the case anyway if there is a chance that one of the messages could actually indicate an unmarshalling error.

jhump (commented):
Thanks for the PR! Sorry it took me so long to leave feedback.

internal/resources/webform/webform-sample.css
theMPatel: updated!

internal/resources/webform/webform.js
theMPatel: I updated it per your suggestion, what's the right way to test multiple messages? Want to test that the responses show up correctly but I don't see a clear way to specify more than one request message

jhump: The test server in this repo has all kinds of crazy stuff in its RPC interface, purely for testing grpcui. So it includes streams, both client and server, as well as request fields of all possible types. So you could do this to test:

```bash
go install github.com/fullstorydev/grpcui/testing/cmd/testsvr
testsvr -port 12345 &
grpcui -plaintext localhost:12345
```

Pick the "DownloadMany" method and then find the request field `numbers` > `needed_num_1`. Set that to the number of response messages you want, and the server will reply with that many messages in the response stream.

internal/resources/webform/webform-template.html
```diff
@@ -49,6 +49,7 @@ <h3>Response Headers</h3>
       <h3>Response Data</h3>
       <div id="grpc-response-req-stats"></div>
       <div id="grpc-response-data"></div>
+
```
jhump, line 52: nit: let's undo this whitespace-only change

jhump (commented):
This looks good to me. Just one little nit remaining. Let me know after you get a chance to test out the multiple-responses scenario.

internal/resources/webform/webform.js
theMPatel: Ok! I did what you asked, here's the result. The one thing I couldn't figure out how to do was to have the `testsvr` return an error for one of the messages. The `DownloadMany` endpoint streams the number of messages in the request back to the client but I can't tell how `isError` is actually populated. If the stream closes due to an error, then it looks like the top level `error` field is set and the stream interface doesn't seem to expose a way to send back an error, rather you're expected to include any relevant errors in the message return type of the stream?

Screen Shot 2020-12-21 at 11 11 48 AM

internal/resources/webform/webform-template.html
theMPatel: done!

internal/resources/webform/webform.js
jhump: > The one thing I couldn't figure out how to do was to have the testsvr return an error for one of the messages.

Yep, that is intentional: an error of this form should only happen if something is awry -- like a malformed response or misbehaving server.

It would be _possible_ to hack this, by not using the normal protoc-generated server interface and implementing the service at a _much_ lower level that gives direct control over the bytes to send back (so we could intentionally send back garbage). But I don't think that is worth the level of effort.

jhump (approved):
Nice!

theMPatel:
I tried to kick CI with a force push but travis seems to be stuck on `Travis CI - Pull Request `. wdyt?

jhump:
@theMPatel, this is Travis CI moving away from free/opensource builds :(  Sadly, we probably need to migrate all of our OSS repos to our CircleCI account since I don't think we want to pay two different CI providers (and Travis is recently _way_ more limited for free accounts). In the meantime, I'll pull this down and run all of the CI checks and, if it LG, I can merge it.

jhump:
Manually ran all CI checks. Looks good

theMPatel:
thanks for working with me on this!! I'll work to get this integrated into `FS`

jhump:
@theMPatel, our internal version is based on https://github.com/fullstorydev/grpcui/pull/79. So let me rebase that PR with your latest changes.

mrinal-sharma:
@jhump @thempatel I want to use this feature, how can I use this?

PR #103, opened 2020-12-14
Add -base-path option to configure base path.

colega:
As documented in /standalone, `http.StripPrefix` can be used to serve requests on paths different from root.

Add this as an option to the binary to make routing to different grpcui instances easier.

### DIFF
--- a/cmd/grpcui/grpcui.go
+++ b/cmd/grpcui/grpcui.go
@@ -110,6 +110,10 @@ var (
 		The port on which the web UI is exposed.`))
 	bind = flags.String("bind", "127.0.0.1", prettify(`
 		The address on which the web UI is exposed.`))
+	basePath = flags.String("base-path", "/", prettify(`
+		The path on which the web UI is exposed.
+		Defaults to slash ("/"), which is the root of the server.
+		Example: "/debug/grpcui".`))
 	services multiString
 	methods  multiString
 )
@@ -259,6 +263,9 @@ func main() {
 	}
 	if len(importPaths) > 0 && len(protoFiles) == 0 {
 		warn("The -import-path argument is not used unless -proto files are used.")
+	}
+	if len(*basePath) == 0 || (*basePath)[0] != '/' {
+		fail(nil, "The -base-path should be non-empty and begin with a slash (\"/\")")
 	}
 
 	configs, err := computeSvcConfigs()
@@ -444,13 +451,20 @@ func main() {
 			logInfof("%s %s %s %d %dms %dbytes", r.RemoteAddr, r.Method, r.RequestURI, cs.code, millis, cs.size)
 		})
 	}
+	if *basePath != "/" {
+		handler = http.StripPrefix(*basePath, handler)
+	}
 
 	listener, err := net.Listen("tcp", fmt.Sprintf("%s:%d", *bind, *port))
 	if err != nil {
 		fail(err, "Failed to listen on port %d", *port)
 	}
 
-	url := fmt.Sprintf("http://%s:%d/", *bind, listener.Addr().(*net.TCPAddr).Port)
+	path := *basePath
+	if len(path) > 1 {
+		path += "/"
+	}
+	url := fmt.Sprintf("http://%s:%d%s", *bind, listener.Addr().(*net.TCPAddr).Port, path)
 	fmt.Printf("gRPC Web UI available at %s\n", url)
 
 	if *openBrowser {


colega:
I just saw that this duplicates a previous effort in https://github.com/fullstorydev/grpcui/pull/54

Since this one has no conflicts with current master, I'll leave it open.

This is tested to work properly and it's actually what we're using in our development environment.

jonaskello:
Yes, please add this :-). As I mentioned in #54  I tried to use it to serve grpcui using the official docker image under a sub-path of my service in k8s with ingress nginx but it did not work, just got "404 page not found". I does however work if I service from "/". Would be super useful for serving under "my.service.com/grpcui".

jonaskello:
To give some more background check [this link](https://docs.giantswarm.io/guides/advanced-ingress-configuration/#path-based-fanout):

> You can route an Ingress to different Services based on the path:
> Note: Your applications need to be capable of running on a non-root path either by default or by setting the base path in their configuration.

jonaskello:
Btw, `nginx.ingress.kubernetes.io/rewrite-target:` does not work in this case other than for the inital html document. Grpcui will turn around to look for every other file directly on the root, eg. `/grpc-logo.png`, `/jquery-3.4.1.min.js` etc,

jonaskello:
I managed to solve my particular problem with ingress nginx in k8s, see [this comment](https://github.com/fullstorydev/grpcui/pull/54#issuecomment-752164819).

colega:
@jhump should I just close this?

cmd/grpcui/grpcui.go
```diff
@@ -260,6 +264,9 @@ func main() {
 	if len(importPaths) > 0 && len(protoFiles) == 0 {
 		warn("The -import-path argument is not used unless -proto files are used.")
 	}
+	if len(*basePath) == 0 || (*basePath)[0] != '/' {
+		fail(nil, "The -base-path should be non-empty and begin with a slash (\"/\")")
```
jhump, line 268:
```suggestion
	if !strings.HasPrefix("/") {
		fail(nil, `The -base-path must begin with a slash ("/")`)
```

cmd/grpcui/grpcui.go
```diff
@@ -444,13 +451,20 @@ func main() {
...
+	}
 
 	listener, err := net.Listen("tcp", fmt.Sprintf("%s:%d", *bind, *port))
 	if err != nil {
 		fail(err, "Failed to listen on port %d", *port)
 	}
 
-	url := fmt.Sprintf("http://%s:%d/", *bind, listener.Addr().(*net.TCPAddr).Port)
+	path := *basePath
+	if len(path) > 1 {
```
jhump, line 464:
```suggestion
	if !strings.HasSuffix(path, "/") {
```

cmd/grpcui/grpcui.go
```diff
@@ -444,13 +451,20 @@ func main() {
 			logInfof("%s %s %s %d %dms %dbytes", r.RemoteAddr, r.Method, r.RequestURI, cs.code, millis, cs.size)
 		})
 	}
+	if *basePath != "/" {
+		handler = http.StripPrefix(*basePath, handler)
```
jhump, line 455: IIRC, this can act funky if the given prefix ends in "/". For example, if we used "/grpcui/" as the prefix, then "/grpcui" results in a 404 and even "/grpcui/" is weird because it strips the leading "/" from the URI (which I think means it won't correctly match any URIs in the underlying mux).

So I think we need to remove any trailing slash.

```suggestion
		handler = http.StripPrefix(strings.TrimSuffix(*basePath, "/"), handler)
```

jhump (commented):
@colega, sorry this PR was neglected for so long. I was initially hesitant because I thought using external configuration (like in a proxy) would be better. But I've come around to it.

So I left some comments.

cmd/grpcui/grpcui.go
colega: Oh this is definitely better. Just missing the first param:
```suggestion
	if !strings.HasPrefix(*basePath, "/") {
		fail(nil, `The -base-path must begin with a slash ("/")`)
```

cmd/grpcui/grpcui.go
```diff
@@ -444,13 +451,20 @@ func main() {
 			logInfof("%s %s %s %d %dms %dbytes", r.RemoteAddr, r.Method, r.RequestURI, cs.code, millis, cs.size)
 		})
 	}
+	if *basePath != "/" {
+		handler = http.StripPrefix(strings.TrimSuffix(*basePath, "/"), handler)
```
jhump, line 455: I checked this out and was playing with it. Another quirk of `StripPrefix` is that if I fail to include the trailing slash, after stripping, it sees a URI of `""`. Since the underlying handler is a mux that has `"/"` registered, it redirects to `"/"` -- without the prefix. So it's a bit odd that knows to do a redirect (but from `"/base-path"` to `"/"`) but still ultimately 404s.

It's a bit of a shame that the `StripPrefix` handler is so unsophisticated, because the result is that we need to do just a little bit more here:
```suggestion
		withoutSlash := strings.TrimSuffix(*basePath, "/")
		mux := http.NewServeMux()
		// the mux will correctly redirect the bare path (without trailing slash)
		mux.Handle(withoutSlash + "/", http.StripPrefix(withoutSlash, handler))
		handler = mux
``````

jhump (commented):
The code looks great. But after toying around locally, I have one last suggestion.

PR #109, opened 2021-03-02
Retry connection to server

### LINKED ISSUE #101: retry upstream connection
I'm using grpcui in a docker-compose app. I've got a bit of a race condition where the upstream gRPC server often doesn't start before grpcui attempts to connect to it. I'm using docker-compose `depends_on`, but this doesn't work because the container with the gRPC service reports ready when the server process has started, and not when the server is actually ready to receive connections. There's normally a few seconds in it.

Grpcui is making one attempt to connect to the upstream service, and then exiting with an error code. What i'd really like is a way to specify that it should retry the upstream connection. Is there a clean way to do this?

troyswanson:
Introduces the ability for grpcui to retry connecting to a server.

When the `retry-max-attempts` flag is supplied (and the value is greater than 0), the initial connection routine will attempt to retry connecting to the server, and exponentially back off the amount of time between connection attempts (from a minimum of 2 seconds to a maximum of 8 seconds).

This makes using grpcui as a sidecar to gRPC service development in docker-compose extremely easy.

docker-compose.yaml
```
version: "3.8"
services:
  grpc:
    build:
      context: .
      dockerfile: ./Dockerfile
      target: dev-runner
    env_file: .env
    volumes:
      - .:/usr/src/app
    ports:
      - 4045:4045
  ui:
    image: fullstorydev/grpcui:v1.1.0-3-gfe2bd93-dirty
    command: -plaintext -open-browser=false -retry-max-attempts=5 grpc:4045
    ports:
      - 8080:8080
```

Terminal output
```
~/Projects/testproj $ docker-compose up
Recreating testproj_ui_1 ... done
Starting testproj_grpc_1 ... done
Attaching to testproj_grpc_1, testproj_ui_1
ui_1    | Unable to reach host "grpc:4045"
ui_1    | Reconnecting in 2s
grpc_1  | [nodemon] 2.0.7
grpc_1  | [nodemon] to restart at any time, enter `rs`
grpc_1  | [nodemon] watching path(s): *.*
grpc_1  | [nodemon] watching extensions: py
grpc_1  | [nodemon] starting `python -m testproj`
grpc_1  | server starting on 0.0.0.0:4045
ui_1    | Retry attempt 1 of 5
ui_1    | gRPC Web UI available at http://0.0.0.0:8080/
```

This closes #101

### DIFF
--- a/cmd/grpcui/grpcui.go
+++ b/cmd/grpcui/grpcui.go
@@ -23,6 +23,7 @@ import (
 	"github.com/fullstorydev/grpcurl"
 	"github.com/jhump/protoreflect/desc"
 	"github.com/jhump/protoreflect/grpcreflect"
+	"github.com/jpillora/backoff"
 	"github.com/pkg/browser"
 	"golang.org/x/crypto/ssh/terminal"
 	"golang.org/x/net/context"
@@ -84,6 +85,11 @@ var (
 		probe is sent. If the connection remains idle and no keepalive response
 		is received for this same period then the connection is closed and the
 		operation fails.`))
+	retryMaxAttempts = flags.Int("retry-max-attempts", 0, prettify(`
+		The maximum number of times to attempt to reconnect to the gRPC server.
+		grpcui will exponentially back off the time between connection attempts
+		as it is unable to reach the server. Useful for running grpcui as a
+		sidecar to gRPC service development. Defaults to 0.`))
 	maxTime = flags.Float64("max-time", 0, prettify(`
 		The maximum total time a single RPC invocation is allowed to take, in
 		seconds.`))
@@ -233,6 +239,9 @@ func main() {
 	if *maxMsgSz < 0 {
 		fail(nil, "The -max-msg-sz argument must not be negative.")
 	}
+	if *retryMaxAttempts < 0 {
+		fail(nil, "The -retry-max-attempts argument must not be negative.")
+	}
 	if *plaintext && *insecure {
 		fail(nil, "The -plaintext and -insecure arguments are mutually exclusive.")
 	}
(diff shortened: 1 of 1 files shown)


codebutler:
can someone please review this? @jameremo @jhump

jhump (commented):
Sorry that this has sat with no feedback for so long. I looked at it when it was first opened and remember typing up something. But I guess I never published those comments (and GitHub no longer shows them as draft comments either...).

The feedback I thought I already gave was that the `grpcurl.BlockingDial` function _already_ does retries and exponential backoff internally -- that's functionality in the underlying gRPC runtime's `Dial` function.

So this means we'd now be doing nested retries (which means a TCP dump would likely show the actual number of attempts to be higher than the given `-retry-max-attempts` value).

I'm going to guess that the real underlying problem might be that "connection refused" or "no route to host" type errors are considered permanent by the gRPC runtime, so [this dial option](https://github.com/fullstorydev/grpcurl/blob/v1.8.1/grpcurl.go#L640) causes the dial operation to fail fast instead of waiting for the remote server to become available.

I think the better fix would be an option like `-connect-fail-fast` that defaults to true and, when false, the call to `grpcurl.BlockingDial` [here](https://github.com/fullstorydev/grpcui/blob/v1.1.0/cmd/grpcui/grpcui.go#L328) should include an explicit `grpc.FailOnNonTempDialError(false)` dial option.

However, that is going to first require a fix over in the grpcurl repo, since the current formulation will always ignore/override the explicitly provided dial option.

troyswanson:
@jhump Thanks for the review! I think I'll close this PR since the implementation isn't what you were looking for, and watch the grpcurl project for a way to configure that `grpc.FailOnNonTempDialError` option.

PR #114, opened 2021-04-22
smarter handling of null values in JSON raw request

jhump:
There was logic that attempted to handle null values as elements in arrays (repeated fields) or values in maps, but it wasn't quite right. In most cases, building the form would complain about a value of the wrong type

This PR augments that behavior so it works in all cases, such as when the `null` is an element in a JSON array or value in a JSON object that corresponds to a map field instead of a message.


internal/resources/webform/webform.js
```diff
@@ -407,11 +407,27 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug, headers)
         input.childKeys = [];
 
         var arrayVal = [];
-        for (k in value) {
+        for (var k in value) {
             if (value.hasOwnProperty(k)) {
+                var keyField = mapEntryFields[0];
+                var valField = mapEntryFields[1];
```
claygoddard, line 413: nit: Move these vars up even higher? They are also used line 390.

internal/resources/webform/webform.js
```diff
@@ -407,11 +407,27 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug, headers)
...
+                if (keyField.type !== "string") {
+                    // k should be a number or a bool
+                    try {
+                        k = JSON.parse(k);
+                        // success!
+                    } catch {
+                        // failure! leave k alone
+                    }
+                }
+                var v = value[k];
```
claygoddard, line 423: Does this actually work? Converting k to a different type and then doing a lookup in the map that originally had string keys? I guess so in Javascript..

You also could just do `const [k, v] of Object.entries(value)` if you wanted.

claygoddard (approved):
Minor comments. LGTM

internal/resources/webform/webform.js
jhump: It does work. But I can see why it's confusing; I can move this line above the conversion of `k`.

Regarding your suggestion, neat, I didn't realize that was a thing -- maybe EcmaScript 6? As you can see, my JS knowledge is for ES 5 (I haven't done much JS since the new stuff was introduced....).

internal/resources/webform/webform.js
jhump: Roger

jhump:
I'm going to hold off on merging this. I realize that, if the upstream projects insist that null values inside an array or map value are invalid, then maybe grpcui should not support that. Much of the changes herein are still valid, but I would probably just amend the code in `addArrayToForm` and `addMapToForm` to explicitly disallow a null value and provide a better error message.

Issue #116, opened 2021-05-03
Time on history tab seems to include rendering time


kodedylf:
Calls to services that return a lot of data take a lot longer in grpcui than in grpcurl. I suspect the time is spent rendering the html.

The time displayed on the history tab reflects the time the call takes in grpcui, including the rendering. It would be nice if it only displayed the time it took to make the service call.

kodedylf:
I think I know how to fix this. I tried forking the repo, but it seems golang is very resistant to forks and I can't build from the fork. I don't have any golang experience, so maybe I'm just dumb. The fix is in Javascript.

jhump:
As long as you are using Go 1.13 or higher and you do _not_ have `GO111MODULE=off` in your environment, it should be fine to fork: you just have to run the Go tool from _inside_ the module (in the repo's root directory, with the `go.mod` file, or anywhere therein).

The issue you're encountering might instead be how the JS code is packaged into the Go binary/library: it requires running a separate step to _generate_ Go code from the JS. You can do this via `make generate` inside the repo's root directory. Then you'll have to restart any running `grpcui` process (no auto-load of assets in development mode, sorry).

PR #123, opened 2021-05-18
Get duration right after call completes

kodedylf:
Capture the duration of the call right after it completes. Since the response is rendered as JSON now rather than HTML, this is less of an issue than it was before. But capturing it as quickly as possible is better.


kodedylf:
This should be linked to issue #116

jhump:
@kodedylf, does this resolve the latency discrepancy you see between grpcurl and grpcui? Another way to handle this and provide better timing accuracy would be to have the server time the RPC and include the duration in the response.

kodedylf:
@jhump, having the server time it would be even better. Tbh., I can't see that big of a difference anymore since the rendering is much faster now with JSON.
Feel free to reject the PR and close the issue if you want. The switch to JSON has solved my main issue, so I'm happy :)

PR #125, opened 2021-06-09
fix: timeFormat should support +01:00

wolfogre:


jhump (commented):
Did you test this? I don't think the fix as is can work. You have to also re-generate code (`make generate`) in order to bring updated JS code into the resulting Go binary.

wolfogre:
Sorry about that, I re-generated and tested the code, it works as expected now.

image

jhump (approved):
Thanks for the fix!

PR #130, opened 2021-06-30
update dependencies

### LINKED ISSUE #129: Failed to process source files with proto3 optionals
Getting the following error when trying to run grpcui with -proto and protos that include proto3 optionals: 

`Failed to process proto source files.: could not parse given files: Example.proto:58:5: field ExampleResponse.message: field has label LABEL_OPTIONAL, but proto3 must omit labels other than 'repeated'`

This appears to be due to the dependency on grpcurl 1.7.0, which does not support the new proto3 optionals

Replacing in a local build with grpcurl 1.8.1 appears to fix the issue, but fires off a bunch of deprecation warnings on make

jhump:
This also re-generates proto/grpc stuff using the new code generators and fixes deprecations (since the new version of the proto runtime has deprecated a lot of the old packages).

It's not as bad as the diff summary looks since the great majority of changes are in generated files.

Fixes #129.


jameremo (approved):
LGTM!

PR #131, opened 2021-07-09
standalone: support proto files as reflection alternative

data-enabler:
For cases where server reflection doesn't work, being able to simply
provide the proto files is a useful alternative.

### DIFF
--- a/files.go
+++ b/files.go
@@ -2,6 +2,7 @@ package grpcui
 
 import (
 	"github.com/jhump/protoreflect/desc"
+	"github.com/jhump/protoreflect/desc/protoparse"
 	"github.com/jhump/protoreflect/grpcreflect"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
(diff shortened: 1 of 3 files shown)


data-enabler:
Just realized that all the APIs involved are public, and from the documentation it sounds like they explicitly support people creating their own handlers using those methods. So unless we think this would be a useful upstream contribution, we could probably avoid forking the library at all.

jhump:
I don't think these new utility functions carry their weight. The command-line tool already supports proto source files, because there is already adequate support in the underlying `grpcurl` library, for creating a `DescriptorSource` from source files.

You could just define these small helpers in your own client of `grpcui`.

data-enabler:
> I don't think these new utility functions carry their weight. The command-line tool already supports proto source files, because there is already adequate support in the underlying `grpcurl` library, for creating a `DescriptorSource` from source files.
> 
> You could just define these small helpers in your own client of `grpcui`.

Oh whoops! Intended to open a PR against our own fork and, didn't realize GitHub would create it against the fullstory repo by default. I definitely agree that that these can just be helpers in our own code; I'm about to do that right now.

Sorry for the mixup, but thanks for the comments 👍

PR #133, opened 2021-07-22
Ability to have grpcui not fail-fast on connection errors

### LINKED ISSUE #101: retry upstream connection

jhump:
Fixes #101.

I had originally hoped to address this with changes in the underlying `grpcurl.BlockingDial` function. And I even submitted a change to that function. However, after trying it out, I discovered that the change did not actually work -- there are other issues with `BlockingDial` that cause it to fail fast, and they are not easy to resolve, especially in a backwards-compatible way.

So I instead had to fork the dialing logic since what we want here is just completely different from `BlockingDial` -- it is very much focused on the interactive use case, and has several places that cause it to intentionally fail-fast (with as good of an error message as possible, since the error returned from `grpc.DialContext` can be bad and can obscure underlying connection error messages).

So this adds a new flag to `grpcui` so users can do `-connect-fail-fast=false`. In this mode, it does not use `grpcurl.BlockingDial` at all. Instead it uses its own dial logic that intentionally never fails fast, so it will rely on the underlying `grpc.Dial` capability to perform retries and backoff. It still does some decoration around the call to `grpc.Dial`, just to get the best error message possible (since, as mentioned above, the gRPC runtime can return bad error messages).

The first commit is not strictly necessary. That was when I was trying out the changes to `grpcurl.BlockingDial`, before realizing they were inadequate. But I left the commit here since it seems reasonable to be up-to-date with the latest releases for these dependencies.

### DIFF
--- a/cmd/grpcui/grpcui.go
+++ b/cmd/grpcui/grpcui.go
@@ -759,7 +759,7 @@ func dial(ctx context.Context, network, addr string, creds credentials.Transport
 	if err := errCreds.err(); err != nil {
 		return nil, err
 	}
-	// otherwise, use the error the dialer last error
+	// otherwise, use the error the dialer last observed
 	if err := dialer.err(); err != nil {
 		return nil, err
 	}
--- a/cmd/grpcui/grpcui.go
+++ b/cmd/grpcui/grpcui.go
@@ -18,6 +18,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/fullstorydev/grpcurl"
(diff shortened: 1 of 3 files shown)


cmd/grpcui/grpcui.go
```diff
@@ -718,3 +725,91 @@ func logInfof(format string, args ...interface{}) {
...
+		c.lastErr = err
+	}
+	return conn, err
+}
+
+func (c *errTrackingDialer) err() error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.lastErr
+}
```
dragonsinth, line 815: Wow... it's kind of shocking that this level of invasive wrapping is needed just to get a decent error back.

At any rate, cmd/grpcui seems like a distinctly weird place for this logic to live...  at least if there were `grpcurl.DialContextWithBetterErrors`, it could be reused.  I'll approve, but I think your TODO is prescient.

dragonsinth (approved):

PR #141, opened 2021-10-07
fix errors related to local storage quota

### LINKED ISSUE #124: Error when storing large responses in history
When calling a service that returns a large response, gRPCui fails when storing the call in the history. No error message is shown and neither is the response.

The error in the browser console is
```
Uncaught DOMException: Failed to execute 'setItem' on 'Storage': Setting the value of 'grpcui-history-localhost:60551-host.docker.internal:8017' exceeded the quota.
    at onHistoryChange (http://localhost:60551/grpc-web-form.js:2431:22)
    at addHistory (http://localhost:60551/grpc-web-form.js:2445:9)
    at Object.<anonymous> (http://localhost:60551/grpc-web-form.js:2199:21)
    at c (http://localhost:60551/jquery-3.4.1.min.js:2:27742)
    at Object.fireWith [as resolveWith] (http://localhost:60551/jquery-3.4.1.min.js:2:28487)
    at l (http://localhost:60551/jquery-3.4.1.min.js:2:78789)
    at XMLHttpRequest.<anonymous> (http://localhost:60551/jquery-3.4.1.min.js:2:81117)
```

jhump:
Specifically:
* Use _less_ local storage: don't store entire RPC response since it's not even used. Now we only store the minimum response info for rendering item in history.
* We limit the data we'll write to a single key to 1mb. Most grpcui instances will have only one key, so this should keep us well under normal quota (which is 5mb).
* Even after above limit, if we still hit quota exception: don't blow up! Instead, log the error and keep going.

This also includes a couple of other minor fixes for things I identified while testing out the history changes/fixes (see comments below for more details).

Resolves #124.


dragonsinth:
Chris days he could, just not immediately

jhump:
> I'm not sure it's fair to ask Chris to keep contributing.

Agreed. I also don't like the idea of having to wait on external contributors, who can't be realistically expected to prioritize this, in order to land fixes to stuff that we own.

Issue #150, opened 2021-12-15
Disable connect fail fast, is it possible?


sbchisholm:
With the `-connect-fail-fast` defaulting to `true` from what I can tell there isn't a way to pass the flag so that it becomes `false`. If there isn't a way to make this `false` perhaps we could rename it to `-no-connect-fail-fast` and have it `false` by default.

```
	connectFailFast = flags.Bool("connect-fail-fast", true, prettify(`
		If true, non-temporary errors (such as "connection refused" during
		initial connection will cause the program to immediately abort. This
		is the default and is appropriate for interactive uses of grpcui. But
		long-lived server use (like as a sidecar to a gRPC server) will prefer
		to set this to false for more robust startup.`))
```

jhump:
> from what I can tell there isn't a way to pass the flag so that it becomes false

This is not true: `-connect-fail-fast=false`

jhump:
https://pkg.go.dev/flag#hdr-Command_line_flag_syntax

sbchisholm:
Ah, thanks @jhump.  I had tried `-connect-fail-fast false`, which didn't work but I see they mention this in their docs:

> -flag
> -flag=x
> -flag x  // non-boolean flags only

PR #151, opened 2021-12-15
Flip logic of -connect-fail-fast to -no-connect...

sbchisholm:
Proposed fix for #150.


sbchisholm:
Not necessary as the flag can be set to `false` by using the `-flag=false` syntax, I assumed this wasn't working because it wasn't working with a space (`-flag false`).

Issue #152, opened 2021-12-20
is type google.protobuf.Value supported?


hyperyuri:
I cant set the value when the field is google.protobuf.Value
![image](https://user-images.githubusercontent.com/54409166/146820010-ebf9500a-5ed6-4859-a4fd-77d85cb448cd.png)

This type is generic like interface{} or any in another languages

jhump:
This is a bug. `grpcui` does have functionality for supporting this type, by providing a text area into which you can enter any valid JSON. The bug is that it should probably render an unset value as a disabled empty text area, so that when you click the checkbox to add a valid, it can correctly enable the text area. Another possible solution will be to have it replace the _unset_ label with the text area when you check the box.

hyperyuri:
how can i help to solve this bug?

hyperyuri:
I am thinking to make protobuf.Value be like protobuf.Any is it good a way to solve the problem?

jhump:
So I now see that `google.protobuf.Struct` works correctly -- it shows _unset_ when the field is absent. And when you check the box, it replaces that label with a text area. And `google.protobuf.Value` should work _exactly_ the same -- the only difference between the two types is that `Value` can be _any_ kind of value where as `Struct` must be an object/map.

So if you are comfortable debugging JS code in the browser, you might step through things to figure out what is happening differently between `Struct` and `Value` when that box is checked.

The `unset` label comes from [here](https://github.com/fullstorydev/grpcui/blob/master/internal/resources/webform/webform.js#L1772). And the code to populate the form with a JSON text area is [here](https://github.com/fullstorydev/grpcui/blob/master/internal/resources/webform/webform.js#L1593). The code that reacts to the checkbox changing, and is supposed to swap in an element, replacing _unset_, is [here](https://github.com/fullstorydev/grpcui/blob/master/internal/resources/webform/webform.js#L728).

Lastly, in this repo in `./testing/cmd/testsvr` is a simple server with example methods that can exercise every possible type of input. So, for this case, it exposes a `DoManyThings` method whose request has a field named `wk` (which contains lots of fields for testing the form entry for well-known types, including `google.protobuf.Value`). Therein is a field named `optional_value` for testing precisely this case.

jhump:
> I am thinking to make protobuf.Value be like protobuf.Any is it good a way to solve the problem?

No, you don't need to do that. Value already has correct support. This is just a bug somewhere in the code that handles the checkbox being checked.

[...]

PR #153, opened 2021-12-21
fix: google.protobuf.Value showing unset

hyperyuri:
About issue #152

On server the value can be:

image

@jhump


internal/resources/webform/webform.js
```diff
@@ -1923,7 +1923,7 @@ window.initGRPCForm = function(services, invokeURI, metadataURI, debug, headers)
             case "google.protobuf.BoolValue":
                 return false;
             case "google.protobuf.Value":
-                return null;
+                return "";
```
jhump, line 1926: I think `null` is a better default than empty string (as would be `{}`). Also, `null` is an acceptable value for a `google.protobuf.Value`. So perhaps the real issue is somewhere else, not properly handling the `null` value?

hyperyuri: yes, i will debug to see where it's not handling the null value

hyperyuri: we can made an if inside `addOptionalMessage` to check if `fld` is `google.protobuf.Value` to then return `addElementToForm(schema, container, parent, pathLen, value, allowMissing, fld);`

hyperyuri: if you agree with this solution or the other one, please tell me, than i will fix

hyperyuri: i already test it with the if and it works

jhump (approved):
LGTM

jhump (commented):
Before I can merge this, you need to run `make generate`. This JS file is actually incorporated into the Go package in the generated file `internal/resources/webform/bindata.go`. And running `make generate` should update that generated file with this one line change.

PR #156, opened 2022-01-23
use go:embed to bind data

xwjdsh:
Simplify code.


jhump:
@xwjdsh, thanks for the patch!

We intentionally support Go runtimes older than G 1.17. After 1.18 is GA (which should be in the coming month), maybe we can re-assess since, IIUC, the Go team will drop support for 1.16 at that point (based on their [release policy](https://go.dev/doc/devel/release#policy)).

In the meantime, let's leave this pull request open and consider merging it after that.

xwjdsh:
@jhump actually the embed feature added in go1.16.

spatecon (approved):

dragonsinth:
This would need a rebase for conflict

xwjdsh:
@dragonsinth Updated.

dragonsinth:
https://github.com/fullstorydev/grpcui/pull/156/files#diff-288b5fde448529a95912cdc56f6eddfbb5baaff11e11bf53c3d1e3efed7c91b5R57

maybe `// Add embedded resources bundled in standalone package TOC`?

go1. 17 (released 2021-08-16)
@jhump seems reasonable to support go 17, 18, 19 to me.

jhump:
> @jhump seems reasonable to support go 17, 18, 19 to me.

Sure, seems fine to me. I had commented almost a year ago, before 1.18 was released.

dragonsinth (approved):

Issue #127, opened 2021-06-22
unable to send google.protobuf.Timestamp in Safari


Gurpartap:
<img width="665" alt="Screenshot 2021-06-23 at 03 17 19" src="https://user-images.githubusercontent.com/39792/123003929-86e96380-d3d1-11eb-84e9-4640b7b2057e.png">

It fails validation in UI, and sends epoch 0 time on server.

Works in Chrome though.

Tried both `github.com/fullstorydev/grpcui v1.1.0` and `github.com/fullstorydev/grpcui v1.1.1-0.20210615152741-73d509c2f2ba` (master as of today).

Safari Version 14.1 (16611.1.21.161.6)
macOS 11.3.1 (20E241)

Side question, is there a way to restore the htmlized response structure UI (as was in previous versions)?

jhump:
Fixed in #159.

PR #159, opened 2022-02-10
Fix timestamp on safari

theJenix:
This PR fixes two Timestamp related issues when using grpcui in Safari:
- when you send a message with a Timestamp, and you've changed the date, the sent message contains 0 or 1970-01-01 00:00:00 for the Timestamp field
- when a user edits the time component and tries to Invoke or just tab out of the control, the entered time will always fail validation

(Both are covered in issue #127 )

Safari does not parse Date strings with a space in the middle.  For example `new Date("2022-02-02 12:00:00")` returns a valid date in Chrome, but returns Invalid Date in Safari.  Safari and Chrome both support the same general date time format with a T in the middle, e.g. `new Date("2022-02-02T12:00:00")`, so this PR changes webform.js onDateChange and onTimeChange to build the date time string with a T in between the date and time.


theJenix:
> LGTM. Thanks for the fix!

No prob!  This tool is a life safer, thanks for making it available!

PR #148, opened 2021-11-22
Include protobuf comments in web form

n0v1:
This PR implements the display of explanatory comments from the protobuf file in the web form (#50).

In the current state comments of methods (rpcs) are displayed and comments attached to single fields. This can be easily extended to other comments (e.g. for services, message types, ...). The main question is where and how to display information from the comments. For method descriptions I opted for displaying them directly below the method select, field comments will be displayed when hovering an icon to not overload the form.

This feature is hidden behind boolean option `comments`. Use command line flag `-comments` (or `-comments=true`) to activate it.

Tested with plain proto and descriptor set compiled with `–include_source_info`.

Please review, test, discuss.

Example:

Method comments in protobuf file:
image

Field comments in protobuf file:
image

Result:
image

PR #166, opened 2022-02-14
add descriptions for services, methods, and fields

jhump:
The "descriptions" come in the form of proto source -- showing the service, method, and field in source form (though not necessary exactly as was present in the original source file, but close, including options and whatever comments are preserved in the descriptor).

* Descriptions for the service and method are to the right of the selection drop-downs. They are hidden by default, to minimize the intrusiveness of this change for users that aren't interested in the information. It can be expanded via the arrow next to the service drop-down. Expanding it is "sticky" -- the preference will be remembered across visits.
* Field descriptions are present in the form of tool-tips on the field names. So hovering over the field name will get you the same proto source for the field, including comments and options if present.

*Still to do:* descriptions for oneofs, message, and enum types. This could also be done as tool tips over type names and over the oneof name. However, much of the information is redundant -- oneofs and messages don't need to show their full body since all of the fields are already expanded in the web form. Enums could still show their full contents, as that is the only way for a user to see comments and options on particular enum values.

This supercedes #79 and #148.


jhump:
Examples of what it looks like:
#### Service and method description:
image

#### Field description in tool-tip:
image

go.sum
```diff
@@ -110,8 +110,8 @@ github.com/hashicorp/golang-lru v0.5.0/go.mod h1:/m3WP610KZHVQ1SGc6re/UDhFvYD7pJ
 github.com/hashicorp/golang-lru v0.5.1/go.mod h1:/m3WP610KZHVQ1SGc6re/UDhFvYD7pJ4Ao+sR/qLZy8=
 github.com/ianlancetaylor/demangle v0.0.0-20181102032728-5e5cf60278f6/go.mod h1:aSSvb/t6k1mPoxDqO4vJh6VOCGPwU4O0C2/Eqndh1Sc=
 github.com/jhump/protoreflect v1.10.3/go.mod h1:7GcYQDdMU/O/BBrl/cX6PNHpXh6cenjd8pneu5yW7Tg=
-github.com/jhump/protoreflect v1.10.4-0.20220210001327-6224817bacc5 h1:KIpt99Cgc9Vv3Hjbx/ke8ZQAzcH7AzM7/HxFCumGI7s=
-github.com/jhump/protoreflect v1.10.4-0.20220210001327-6224817bacc5/go.mod h1:U7aMIjN0NWq9swDP7xDdoMfRHb35uiuTd3Z9nFXJf5E=
+github.com/jhump/protoreflect v1.11.0 h1:bvACHUD1Ua/3VxY4aAMpItKMhhwbimlKFJKsLsVgDjU=
```
jhump, line 113: This version includes a fix for printing methods: without it, comments on option statements were never included.

PR #167, opened 2022-02-14
Add the ability to load 'examples' and create examples by saving history

### LINKED ISSUE #94: Support for collections
Is there a plan to add collections, similar to what postman has?

jhump:
This resurrects #95 and adds a couple of commits to bring it over the finish line, including incorporating it into the command-line tool.

Resolves #94.


jhump:
Here's how it looks when examples are given. (These example names aren't descriptive because they come from saving history.)
image

PR #188, opened 2022-06-12
Feature: Bazel Support

sgammon:
Hello Fullstory devs,

Similar to fullstorydev/grpcurl#316, this PR adds support for Bazel, both as an internal build tool, and as an external rules package usable from downstream Bazel projects who want to leverage `grpcui`. I don't know if there is any interest in merging this but I thought I'd propose nonetheless.

#### Changes enclosed

- [x] Add support for downstream Bazel projects (invoke via binary)

<details><summary>(Old: To build/run with Bazel)</summary>

### To build/run with Bazel

This guide is abbreviated, as there is a full one on fullstorydev/grpcurl#316 and they are largely the same. However, here are the commands for easy testing:

```
git clone git@github.com:sgammon/grpcui.git -b bazel && cd ./grpcui
```

**Method 1: Local build/run**
```bash
bazel build //:grpcui
```

```bash
bazel run //:grpcui -- -help
```
</details>

**Running `grpcui` in Bazel projects as a binary**

In the user's `WORKSPACE`:
```starlark
workspace(
  # ...
)

grpcui_version = "494bbc3a99f2e75a43bb47b7c19714b1eb79c4ca"

http_archive(
    name = "com_github_fullstorydev_grpcui",
    sha256 = "640c430f41c657fea9fe84910778dd70dc9a6fe8fd70e5e072972308b79defea",
    strip_prefix = "grpcui-%s" % grpcui_version,
    url = "https://github.com/sgammon/grpcui/archive/%s/archive.tar.gz" % grpcui_version,
)

load("@com_github_fullstorydev_grpcui//:grpcui_repositories.bzl", "grpcui_bindist_repository")

grpcui_bindist_repository(
    name = "grpcui",
    version = "1.3.0",
)
```

Then, they can run it in their own Bazel project easily with:
```
bazel run @grpcui//:grpcui
```

Similar to `grpcurl`, support for method 2 is enclosed for Linux and macOS, although Windows would be relatively easy to add. All architectures are included that `grpcui` is released for.


sgammon:
Hey Scott,

Thanks for your response here. By packaging gRPCurl/gRPCui as a binary distribution, this new style of Bazel invocation is much less invasive for downstream projects.

In fact, downstream Bazel projects don't even need rules_go, because Bazel would just be downloading a binary distribution for their arch and then invoking that. There is no need to build the tools in a user's own env.

RE/internal project Bazel, that's entirely up to you guys as a matter of taste, but this mechanism should make it easier to embrace as a Bazel codebase should you wish to (I.e. with downstream Bazel projects alleviated from a need to build from source, you have wider latitude with tooling w/o user interference)

sgammon:
Since you guys already put out checksums for your releases, mapping in the binaries is actually really easy and obvious. :)

We're loving this style of invocation in our code because it's fast and a very thin layer of abstraction. But it's entirely up to you guys if this is useful or not to others.

grpcui.bzl
```diff
@@ -0,0 +1,12 @@
+"""Bazel macro to run a binary copy of gRPC UI."""
+
+def grpcui(name, descriptor, args = []):
```
dragonsinth, line 3: What's this rule do?

dragonsinth:
Okay, so just to be clear, are we really talking about a few mostly-unrelated things?  Please make sure I'm describing this correctly, but it seems like:

1) The bulk of this PR is around using bazel to build grpcui for our own development.

2) The stuff in `grpcui_repositories.bzl` is basically some bazel glue to make it easier to pull things from https://github.com/fullstorydev/grpcui/releases/tag/v1.3.0 and essentially completely separable from the rest of the PR?

3) Having WORKSPACE and BUILD files in our repo also enables people to pull this repo in through `http_archive` rather than `go_repository`?  But that only seems useful for getting at the macro/rule in #2; I notice that you still need a `go_repository` rule from grpcui -> grpcurl in deps.go.

dragonsinth:
For comparison, mapping grpcui via go.mod -> gazelle update deps allows:

```
bazel build @com_github_fullstorydev_grpcui//cmd/grpcui
bazel run @com_github_fullstorydev_grpcui//cmd/grpcui -- -help
```

As you note, this does require a go toolchain setup.  I imagine you care more about third-party open source projects that want to bundle grpcurl/grpcui and are not themselves natively Go?  I can't imagine this matters all that much for "end user" repos.

sgammon:
@dragonsinth as a user of these tools, my primary motivation is easy and clean invocation from Bazel where we don't have a `rules_go` chain in place, or where adding one, or pinning one, may cause unrelated dependency issues.

we are dispatching from Bazel and calling `grpc_extra_deps()` from `com_google_grpc_grpc`, which registers its own `rules_go` toolchain. by invoking through a pre-built binary, we can avoid any issues there and also avoid the time / remote cache load it takes to build the tools before invoking them.

i figured these tools just hadn't gained Bazel support internally yet, but if you guys have chosen not to build with bazel, that's a different story. because i added it to both repos, i imagined the repos would be built alongside each other and would need to align dependencies. many times these tools are exposed publicly only in part, for instance via a copybara job, so my PR would need to align dependencies against both in order to be mergable in a theorized private monorepo at fullstory.

perhaps that isn't the case and i'm just jaded from contributing to google repos 😄

yes, i do think the changes are separable, and shipping the binary invocation piece could be as simple as just exposing the `bindist.bzl` file and having it consume your `.txt` with the checksums. my goal is definitely not to add a bunch of load to the release cycle, and thank you for responding here.

sgammon:
btw, if you mean "end user" repos as in downstream projects that use Bazel, yes I do think the bindist slice of this change is impactful and worth it, at least for us. i do think there is a growing community of native bazel users, and they happen to use gRPC (unsurprisingly), so `grpcui` / `grpcurl` would be good candidates for native bazel support

if you mean "end user" as in non-bazel projects consuming or invoking these tools, i don't think this change is pertinent and they probably would not notice it.

dragonsinth:
I meant end user bazel repos.  Like, a private org's monorepo built using bazel; as opposed to open source libs that use bazel and depend on grpcui/grpcurl.

So, we definitely _don't_ want to use bazel to build grpcurl/grpcui.  We don't need it to operate on and maintain this repo; it would be an ongoing maintenance burden that would provide very little value to us.  The Go build and Go mod systems are strong and built with a lot of the principles that underlie bazel, while being minimalist and specifically tailored to Go with a closed feature set.  In short, Go works better for building just Go than bazel does... and I say that as one of the folks directly responsible for maintaining our internal bazel monorepo.

I do see (to a limited extent) some utility in being able to use precompiled binaries easily in bazel.  But it's hard for me to imagine a world in which vast numbers of projects distributing binaries include a complicated and sophisticated set of macros/rules to it trivial to fetch the appropriate compiled binary that way.  Some kind of generic rule set that could be pointed at https://github.com/fullstorydev/grpcui/releases (or any of thousands of github projects with binary downloads) seems useful, but every individual project maintaining such a ruleset seems... impractical.

sgammon:
@dragonsinth all entirely understandable, but to be clear, the inclusion of the binaries downstream for end-users in Bazel projects of their own would simply be

```
grpcui_version = "494bbc3a99f2e75a43bb47b7c19714b1eb79c4ca"

http_archive(
    name = "com_github_fullstorydev_grpcui",
    sha256 = "640c430f41c657fea9fe84910778dd70dc9a6fe8fd70e5e072972308b79defea",
    strip_prefix = "grpcui-%s" % grpcui_version,
    url = "https://github.com/sgammon/grpcui/archive/%s/archive.tar.gz" % grpcui_version,
)

load("@com_github_fullstorydev_grpcui//:grpcui_repositories.bzl", "grpcui_bindist_repository")

grpcui_bindist_repository(
    name = "grpcui",
    version = "1.3.0",
)
```

that's it, that's all. `grpc_ui_repositories.bzl` would need to be issued as part of the release process, yes, with the checksums -- but perhaps even that could be downloaded as part of the repository? i don't know.

in any case, all good, i'm happy to close the PR or amend as you see fit. w.r.t. running grpcui/grpcurl in our own Bazel repo, we've got things working fine.

dragonsinth:
QQ: your `url` is set to something ended in `archive.tar.gz`, but on our releases page I see a link ending with `fullstorydev/grpcui/archive/refs/tags/v1.3.0.tar.gz`.  Would this involve producing a separate artifact from the normal release process?  I'm having trouble connecting the pieces.  Or are you talking about just maintaining a `grpcui_repositories.bzl` in-repo and having people reference it as a git external repo?

sgammon:
> `grpcui_repositories.bzl` in-repo and having people reference it as a git external repo

yes, in essence, this. people would use `http_archive` to grab the `grpc_ui_repositories.bzl`. notice i'm using a prefix here which excludes the source code and only makes the `.bzl` file available in the downstream project.

then, consuming the rule from there, they get access to a checksum'd binary to invoke on their arch.

grpcui.bzl
sgammon: the idea was this rule would be a macro to invoke the binary from a user's bazel target, but we can exclude, because this would depend on the rest of the `rules_go` stuff.

grpcui_repositories.bzl
```diff
@@ -0,0 +1,111 @@
...
+    sha = _grpcui_version_checksums["%s_%s_%s" % (version, platform[0], platform[1])]
+
+    urls = [link]
+    ctx.download_and_extract(
+        url = urls,
+        sha256 = sha,
+    )
+
+    ctx.file("BUILD", """exports_files(glob(["**/*"]))""")
+    ctx.file("WORKSPACE", "workspace(name = \"{name}\")".format(name = ctx.name))
```
sgammon, line 100: @dragonsinth this is the key part for bindist -- it's just downloading a binary copy of `grpcui` for the proper arch, and making it available via `exports_files(...)`.

sgammon:
i've cleaned up the Gazelle stuff, and can clean up the rest of it here shortly too if you think it's worth it.

dragonsinth:
So we'd need to what, do the release, and then immediately turn around and update this file on master with the checksums, now that we know them?  I assume we'd actually need to map all the old versions as well, and just grow the file over time?

sgammon:
well, i figured the checksums file was produced in automated form somewhere, and assumed it would just be an addendum to that script, whatever it may be, to produce _this_ artifact, since presumably it would know the checksums at that time?

but these are assumptions and feel free to correct them

i think the Bazel repository could actually be smart enough to download the checksums file and enforce it, rather than needing it expressed literally in code, which could help. it would also still be fine to just make it an input parameter with the version, i really appreciate you guys publishing SHA256's in the first place and when the user picks a version, they already have the checksum right there anyway.

this methodology of distributing binaries is common in the Bazel ecosystem, as far as i can tell, for instance, [rules_nodejs](https://github.com/bazelbuild/rules_nodejs) uses the same strategy to download node/npm/yarn, etc.

grpcui.bzl
sgammon: (or perhaps not? i suppose it could just invoke the binary and pass args, so yes, this would be a Bazel macro to invoke the binary from a target, rather than the command line. should you choose to keep it 😁)

sgammon:
@dragonsinth thanks for the review on this, if there is any future interest, happy to resurface this PR after a rebase.

dragonsinth:
> @dragonsinth thanks for the review on this, if there is any future interest, happy to resurface this PR after a rebase.

👍 thanks for pushing on this, sorry to disappoint!

PR #232, opened 2023-03-22
Open browser in separate goroutine since misbehaving impls can block indefinitely.

### LINKED ISSUE #230: -open-browser=true doesn't work properly on faulty open handlers
* Ubuntu jammy, running on WSLg Windows 22H2 (22621.1413)
* xdg-open 1.1.3
* google-chrome 111.0.5563.64, installed via apt
* grpcui [3cc36d3](https://github.com/fullstorydev/grpcui/commit/3cc36d33c1d2d542e641c88367fc1da29e68dba8), using `go install`

Running grpcui doesn't work and the browser keeps trying connection... until you close the browser! If the browser is closed and another is opened then it starts working.

The default open handler is faulty for google-chrome since it blocks until the browser exits. As consequence `browser.OpenURL()` is a blocking call preventing grpcui process to do its job. A workaround might be to make the call asynchronously.

jhump:
This should fix #230.

### DIFF
--- a/cmd/grpcui/grpcui.go
+++ b/cmd/grpcui/grpcui.go
@@ -706,9 +706,11 @@ func main() {
 	fmt.Printf("gRPC Web UI available at %s\n", url)
 
 	if *openBrowser {
-		if err := browser.OpenURL(url); err != nil {
-			fmt.Fprintf(os.Stderr, "Failed to open browser: %v\n", err)
-		}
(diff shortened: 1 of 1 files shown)


dragonsinth (approved):

jhump:
@dragonsinth, if this LG, mind merging? (I no longer have such power 😝)

PR #238, opened 2023-04-16
bugfix: add verification for item.request.metadata

jtsai-dev:
item.request.metadata will be undefined when using -examples and without setting in json file, it needs to be verified


jhump (approved):

jhump:
LGTM, but I'm no longer a committer, so my approval doesn't allow merging.

dragonsinth (approved):

PR #239, opened 2023-04-17
Add "checkgenerate" make target to CI

jhump:
I was curious about the previously merged PR (#238), noticing that it only updated the JS code and not any Go code where the webform was embedded.

I now see that `go:embed` is used, which is why no explicit Go code had to be changed. Nice!

But while playing with it, I realized that `make generate` wasn't being checked in CI, to make sure that all code was up to date from code generators. So this adds such a check to CI.


Makefile
```diff
@@ -32,12 +34,19 @@ docker:
 	@rm VERSION
 
 .PHONY: generate
-generate:
+generate: .tmp/protoc/bin/protoc
 	@go install google.golang.org/protobuf/cmd/protoc-gen-go@a709e31e5d12
 	@go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.1.0
-	@go install github.com/jhump/protoreflect/desc/sourceinfo/cmd/protoc-gen-gosrcinfo
+	@go install github.com/jhump/protoreflect/desc/sourceinfo/cmd/protoc-gen-gosrcinfo@v1.14.1
```
jhump, line 40: This line wouldn't run without this explicit version. The `v1.14.1` tag for this repo is already in `go.mod`, however there are some transitive deps of this package that are not otherwise used by this repo, so they had no entries in `go.sum`. This gets it working again.

dragonsinth: Explicit deps FTW

download_protoc.sh
```diff
@@ -0,0 +1,32 @@
...
+  PROTOC_ARCH="x86_64"
+fi
+
+PROTOC="${PWD}/.tmp/protoc/bin/protoc"
+
+if [[ "$(${PROTOC} --version 2>/dev/null)" != "libprotoc 3.${PROTOC_VERSION}" ]]; then
+  rm -rf ./.tmp/protoc
+  mkdir -p .tmp/protoc
+  curl -L "https://github.com/google/protobuf/releases/download/v${PROTOC_VERSION}/protoc-${PROTOC_VERSION}-${PROTOC_OS}-${PROTOC_ARCH}.zip" > .tmp/protoc/protoc.zip
+  cd ./.tmp/protoc && unzip protoc.zip && cd ..
```
dragonsinth, line 30: pushd/popd not generally available? just curious

download_protoc.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -e
+
+cd $(dirname $0)
+
+PROTOC_VERSION="22.0"
```
dragonsinth, line 7: Could we pass this value in from the Makefile for clarity?

What version were we using before? I noticed the outputs are changed.

dragonsinth (approved):
LGTM with a couple nits

download_protoc.sh
jhump: done

download_protoc.sh
jhump: done

Makefile
```diff
@@ -89,7 +90,6 @@ errcheck:
 test:
 	go test -race ./...
 
-.tmp/protoc/bin/protoc:
+.tmp/protoc/bin/protoc: ./Makefile ./download_protoc.sh
```
jhump, line 93: I added the deps here, so that the compiler is auto-redownloaded if we change the version (which is in the Makefile) or the download script.

dragonsinth: 👍 nice

dragonsinth (approved):

Issue #3, opened 2018-12-03
Copy as "grpcurl"


jhump:
A slick feature would be to be able to `Copy as 'grpcurl'` -- which would extract all connection info from command-line flags as well as the request metadata and data in the web form and use that to generate an equivalent `grpcurl` command.

Similarly, the ability to paste a `grpcurl` command, and have that auto-populate the metadata and request message data would be slick as well.

abourget:
This is an old one... but well needed. :)  I'd highly appreciate

fleytman:
This very important feature, for fast coomunicete in develop team. Now this feature has wombat https://github.com/rogchap/wombat/pull/51, but with bug https://github.com/rogchap/wombat/issues/50#issuecomment-1236460826

dariusdamalakas:
Hey, @jhump , i have given this a shot. Here's what it looks like.  The `grpcurl` updates on every change, as expected.     Wonder if there is a better placement for this.  

![image](https://user-images.githubusercontent.com/131102/232722426-58256252-5960-4cd6-a724-60a720fcd092.png)

if that looks good, can send a PR for this for review.

jhump:
I won't be able to approve/merge any changes as I'm no longer with FullStory. However, I can provide some input on the above described solution:
1. I wouldn't put it in the UI this way. This is too prominent in the UI for how often this is likely to be used -- just adds noise to the request builder.
2. You might instead just make it a tiny link below the tabbed view (so it can be seen below the form no matter which tab is selected). The link text could be `Copy as "grpcurl"`.
3. An alternative to the above might be changing the "Invoke" button to be a combo drop-down+button where "Invoke" is one of the options (and the default) and another is "Copy as grpcurl".
4. Instead of actually showing the grpcurl statement in the UI, the link or button would store the statement directly in the clipboard (e.g. `navigator.clipboard.writeText(grpcurlStatementText)`.

I'm not sure who at FullStory would be able to provide more guidance on what kind of change would be accepted. I know @dragonsinth has helped me merge some small changes, but I'm not sure who should review something of this nature. I think the biggest concern would be getting the UX right -- so that the feature isn't too prominent/distracting but is still discoverable. I had originally envisioned this being a context menu sort of thing, like right-clicking in the form brings it up, but perhaps that is too subtle/hidden...

dragonsinth:
It could be another Tab.
It could also be another pane in `Raw Request (JSON)`, which we could rename to just `Raw Request`.

[...]

PR #241, opened 2023-04-24
Add "raw request gRPCurl" output

dariusdamalakas:
Implements [#3 Copy as "grpcurl"](https://github.com/fullstorydev/grpcui/issues/3). 

**Example:**
image

**With highlights of the changes parts:**
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,12 @@ run `make install`.
 If you encounter compile errors, you could have out-dated versions of `grpcui`'s
 dependencies. You can update the dependencies by running `make updatedeps`.
 
+### Running without install 
+
+```
+go run ./cmd/grpcui/grpcui.go -plaintext localhost:9019
+```
+
 ## Usage
 The usage doc for the tool explains the numerous options:
 ```shell
@@ -250,4 +256,3 @@ protoc --proto_path=. \
 The `--descriptor_set_out` argument is what tells `protoc` to produce a protoset,
 and the `--include_imports` argument is necessary for the protoset to contain
 everything that `grpcui` needs to process and understand the schema.
-
--- a/internal/resources/webform/webform-sample.css
+++ b/internal/resources/webform/webform-sample.css
@@ -750,7 +750,6 @@ button#grpc-history-clear, button#grpc-history-save {
     margin-bottom: 1em;
 }
 
-
 .grpc-curl-panel {
     background-color: #fcfcfc;
     border: 1px solid #dddddd;
--- a/internal/resources/webform/webform-sample.css
+++ b/internal/resources/webform/webform-sample.css
@@ -746,6 +746,11 @@ button#grpc-history-clear, button#grpc-history-save {
     padding-left: 12px;
 }
 
+.grpc-request-raw-container {
+    margin-bottom: 1em;
+}
+
+
 .grpc-curl-panel {
     background-color: #fcfcfc;
     border: 1px solid #dddddd;
--- a/internal/resources/webform/webform-sample.css
+++ b/internal/resources/webform/webform-sample.css
@@ -747,6 +747,7 @@ button#grpc-history-clear, button#grpc-history-save {
 }
 
 .grpc-request-raw-container {
+    margin-top: 1em;
     margin-bottom: 1em;
 }
 
--- a/internal/resources/webform/webform-sample.css
+++ b/internal/resources/webform/webform-sample.css
@@ -745,3 +745,10 @@ button#grpc-history-clear, button#grpc-history-save {
 #grpc-history-list .history-item-panel .history-detail-metadata td {
     padding-left: 12px;
 }
+
+.grpc-curl-panel {
+    background-color: #fcfcfc;
+    border: 1px solid #dddddd;
+    padding: 1em;
+    margin-bottom: 1em;
+}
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -76,6 +76,12 @@ run `make install`.
 If you encounter compile errors, you could have out-dated versions of `grpcui`'s
 dependencies. You can update the dependencies by running `make updatedeps`.
 
+### Running without install 
+
+```
+go run ./cmd/grpcui/grpcui.go -plaintext localhost:9019
```
dariusdamalakas, line 82: It was not clear for me how to run the app locally, took me 10 minutes or so to figure out.

internal/resources/webform/webform-template.html
```diff
@@ -46,7 +46,7 @@ <h3>Examples</h3>
     <div id="grpc-request-response">
       <ul>
           <li id="#grpc-request-tab-button"><a href="#grpc-request-tab">Request Form</a></li>
-          <li id="#grpc-request-raw-tab-button"><a href="#grpc-request-raw-tab">Raw Request (JSON)</a></li>
+          <li id="#grpc-request-raw-tab-button"><a href="#grpc-request-raw-tab">Raw Request</a></li>
```
dariusdamalakas, line 49: Removing `(JSON)` as this is not adding any useful information

internal/resources/webform/webform-template.html
```diff
@@ -132,6 +137,7 @@ <h3>Response Trailers</h3>
         {"name": "{{.Name}}", "value": "{{.Value}}"},
     {{- end}}
     ];
+    window.target = {{ .Target }};
```
dariusdamalakas, line 140: `Target` is the address of the grpcServer where `grpcui` is connecting to.

internal/resources/webform/webform-template.html
```diff
@@ -70,7 +70,12 @@ <h3>Request Timeout</h3>
         <button class="grpc-invoke" disabled>Invoke</button>
       </div>
       <div class="grpc-tabcontent" id="grpc-request-raw-tab">
+        <h3>Request payload</h3>
         <textarea id="grpc-request-raw-text"></textarea><br/>
+        <h3>gRPC curl</h3>
+        <div id="grpc-curl" class="grpc-curl-panel">
+          <span><pre id="grpc-curl-text"></pre></span>
+        </div>
```
dragonsinth, line 78: This PR is awesome, thank you.  My one question is: why the different constructions (textarea vs. pre)?
Seems like we should use the same setup and formatting for both.  I'm not strongly opinionated on which... if you like the new formatting, you could just make the CSS classname something like `grpc-request-panel` and apply it to both.

dragonsinth: Holy crap... I actually had no idea that the Raw Request box was even editable.....

dragonsinth: In that case, I wonder if we can move the `Invoke` button up above the grpcurl area?

I also wonder if there's a way to make it more obvious that the request box is editable..

dariusdamalakas: @dragonsinth    Like this?  

I am not a fan of this, looks messy. Buttons are usually at the bottom. 


image

dragonsinth: Seems fine to me.  I think it's weirder to separate the editable field from the button that it goes with.

dariusdamalakas: @dragonsinth pushed new changes


image

internal/resources/webform/webform-sample.css
```diff
@@ -745,3 +745,15 @@ button#grpc-history-clear, button#grpc-history-save {
 #grpc-history-list .history-item-panel .history-detail-metadata td {
     padding-left: 12px;
 }
+
+.grpc-request-raw-container {
+    margin-bottom: 1em;
+}
+
+
```
dragonsinth, line 753: nit: extra blank line here

dariusdamalakas:
@dragonsinth thanks.  Removed empty line!

internal/resources/webform/webform-sample.css
dariusdamalakas: Removed. Thanks!

jhump:
First nit: it's called "grpcurl" or "[gRPCurl](https://github.com/fullstorydev/grpcurl#grpcurl)", not "gRPC curl".

Another few cents worth of UX feedback, take it or leave it:

This seems overly prominent in the UI... a bit of an eyesore and possible even confusing. Should users try to paste existing `grpcurl` invocations into the textarea, like they can with JSON? Or is it read-only? While it's possibly interesting information, it is more likely a distraction for most people actually using the "Raw Request" tab.

Following how Chrome exposes similar functionality in its dev tools, I think it would be better to instead add a "Copy as grpcurl" button to each row in the "History" tab.

dragonsinth:
This seems like a reasonable incremental thing.  If some sufficiently motivated person wants to turn this into "copy as gRPCurl" that might be a nice improvment.

dragonsinth:
FYI: I pushed a commit to do 2 things:

1) Adjust the spacing a tad, break up with an HR
2) (javascript) ensure consistency between editing the form and copying as curl by shuffling the code a tad.

dragonsinth:
Screenshot 2023-04-27 at 6 18 17 PM

PR #248, opened 2023-05-30
feat: add headers to grpcurl command

gubtos:
Grpcurl command was missing headers parameters. This PR adds it.

### DIFF
--- a/internal/resources/webform/webform.js
+++ b/internal/resources/webform/webform.js
@@ -2105,7 +2105,22 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
     function updateCurlCommand(requestDataJson) {
         let service = $("#grpc-service").val();
         let method = $("#grpc-method").val();
-        gRPCurlTextArea.html(`<div>grpcurl -plaintext -d '${requestDataJson}' ${window.target} ${service}.${method}</div>`);
+
+        var metadataStr = "";
+        var rows = $("#grpc-request-metadata-form tr");
+        for (var i = 0; i < rows.length; i++) {
+            var cells = $("input", rows[i]);
+            if (cells.length === 0) {
+                continue;
+            }
+            var name = $(cells[0]).val();
+            var val = $(cells[1]).val();
+            if (name !== "") {
+                metadataStr += ` -H "${name}: ${val}"`;
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


internal/resources/webform/webform.js
```diff
@@ -2105,7 +2105,22 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
...
+        var rows = $("#grpc-request-metadata-form tr");
+        for (var i = 0; i < rows.length; i++) {
+            var cells = $("input", rows[i]);
+            if (cells.length === 0) {
+                continue;
+            }
+            var name = $(cells[0]).val();
+            var val = $(cells[1]).val();
+            if (name !== "") {
+                metadataStr += "-H \""+name+": "+ val + "\"";
```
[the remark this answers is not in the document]
gubtos: fixed

dragonsinth (approved):
LGTM!  I'll need to pull the branch and test it tho.

internal/resources/webform/webform.js
```diff
@@ -2105,7 +2105,22 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
...
+            var cells = $("input", rows[i]);
+            if (cells.length === 0) {
+                continue;
+            }
+            var name = $(cells[0]).val();
+            var val = $(cells[1]).val();
+            if (name !== "") {
+                metadataStr += ` -H "${name}: ${val}"`;
+            }
+        }
```
dragonsinth, line 2121: One thing I noticed, changing keys/values/rows in the headers doesn't trigger `updateCurlCommand`.  I'm not sure if it's worth the level of invasiveness that would require tho.  Thoughts?

Maybe we could always just force a re-render when the Raw Request tab is shown instead of the crazy stuff we're currently doing.

dragonsinth: For example, I noticed a bug where on first render, if you load the app (or change the selected method) and click on the raw request tab, the curl command contains an `[object Object]` instead of `{}`.

gubtos: Yes. I agree with you. For example, the updateCurlCommand func receives the entire request instead just the body. I believe this would involve a larger PR of refactoring.  But for now, due to lack of time, I can't do that. Merging this PR would help people here a lot.

dragonsinth: fair enough

PR #250, opened 2023-06-09
Add installation instruction for homebrew

ismaildawoodjee:
Added installation instruction using `homebrew` since its the easiest and fastest, and less prone to errors as well.


dragonsinth (approved):
nice, thank you

PR #266, opened 2023-08-31
Resolve dependabot; go 1.21

dragonsinth:
#260 #265


gpassini (approved):

.circleci/config.yml
```diff
@@ -25,23 +19,29 @@ jobs:
...
     working_directory: ~/repo
     docker:
-      - image: cimg/go:1.20
+      - image: cimg/go:1.21
     steps: *simple_job_steps
 
 workflows:
   pr-build-test:
     jobs:
-      - build-1-17
```
gpassini, line 44: Do we already make it explicit which Go versions we test against?

dragonsinth: We can't actually build against 1.17 anymore, there's a dep in the chain that requires 1.18 so I just bumped it up. 🤷

PR #281, opened 2023-11-02
Adding Emit Defaults Option

James-Bartman:
## Problem Statement

When using gRPC UI, I came across a scenario where I did not want default values emitted to the Response tab. While gRPCurl has an option to enable/disable emit-defaults, gRPC UI does not since the call it makes to gRPCurl hard-codes emit-defaults to true. Thus, there was an inconsistency between these two closely related tools.

## Example Proto File

```Proto
service Greeter {
   rpc SayHello (HelloRequest) returns (HelloReply);
}

message HelloRequest {
   string name = 1;
}

message TwoNums {
   int32 abc = 1;
   int32 xyz = 2;
}

message HelloReply {
   oneof HelloReplies {
      TwoNums nums = 1;
      int32 num = 2;
   }
   string answer = 3;
}
```

## Sample Server Implementation - Written in Go

[server.txt](https://github.com/fullstorydev/grpcui/files/13245072/server.txt)

## gRPC UI Output

```JSON
{
  "nums": {
    "abc": 3,
    "xyz": 3
  },
  "answer": ""
}
```

## gRPCurl Output with `-emit-defaults=false`

```JSON
{
  "nums": {
    "abc": 3,
    "xyz": 3
  }
}
```

## Fix

- Added `emit-defaults` flag to the `cmd/grpcui.go` file.
- Set a global variable in `handlers.go` to accept value of `emit-defaults` flag.
- The default value of this flag is `true`, to preserve existing behavior

```bash
grpcui -h
...
-emit-defaults
        Emit default values for JSON-encoded responses. (default true)
...
```

## gRPC UI Output After Fix with `-emit-defaults=false`

```JSON
{
  "nums": {
    "abc": 3,
    "xyz": 3
  }
}
```


James-Bartman:
Just bumping this but not sure exactly who to notify. @dragonsinth @gpassini

handlers.go
```diff
@@ -27,6 +27,8 @@ import (
 	"github.com/fullstorydev/grpcurl"
 )
 
+var emitDefaults = true
+
```
dragonsinth, line 31: This seems bad... I think you need to do something like:
```go
	result := rpcResult{
		descSource:   descSource,
		emitDefaults: options.EmitDefaults,
		Requests:     &reqStats,
	}
```
around line 160.  Then pass it through as a parameter to toRpcError / responseToJSON

handlers.go
```diff
@@ -539,12 +542,13 @@ type rpcError struct {
...
-	Responses  []rpcResponseElement `json:"responses"`
-	Requests   *rpcRequestStats     `json:"requests"`
-	Trailers   []rpcMetadata        `json:"trailers"`
+	descSource   grpcurl.DescriptorSource
+	Headers      []rpcMetadata        `json:"headers"`
+	Error        *rpcError            `json:"error"`
+	emitDefaults bool
+	Responses    []rpcResponseElement `json:"responses"`
+	Requests     *rpcRequestStats     `json:"requests"`
+	Trailers     []rpcMetadata        `json:"trailers"`
```
dragonsinth, line 551: Sorry to nit pick, but would you mind separately the public/private fields while you're in here?  Something like this would be great:
```go
type rpcResult struct {
	Headers      []rpcMetadata        `json:"headers"`
	Error        *rpcError            `json:"error"`
	Responses    []rpcResponseElement `json:"responses"`
	Requests     *rpcRequestStats     `json:"requests"`
	Trailers     []rpcMetadata        `json:"trailers"`

	descSource   grpcurl.DescriptorSource
	emitDefaults bool
}
```
(you'll have to gofmt after)

dragonsinth (commented):
LGTM with one nit

handlers.go
James-Bartman: Sure thing! I ran gofmt after to clean it up. The only thing is I am now getting this error which I don't think I modified:

```bash
/home/circleci/repo/cmd/grpcui/grpcui.go:708:55: listener.Addr undefined (type net.Listener has no field or method Addr)
```

I though it may have something to do with the bump of `golang.org/x/net` from v0.17.0 to v0.20.0 but all the checks passed on the dependabot PRs.

dragonsinth: Ohh... that is super weird.  Doesn't seem like that should be related to your PR :/

dragonsinth: Looks like it passed on re-run 🤷

James-Bartman: Awesome! Always nice to have things magically fix themselves. Everything is ready to go on my end then.

dragonsinth (approved):

PR #292, opened 2024-01-05
add new raw header

aksanart:
i'have problem with adding bunch of headers so i pr this raw header

add header raw
add header raw2

### DIFF
--- a/internal/resources/webform/webform-sample.css
+++ b/internal/resources/webform/webform-sample.css
@@ -202,6 +202,10 @@ textarea#grpc-request-raw-text {
     height: 200px;
 }
 
+textarea#grpc-request-raw-header {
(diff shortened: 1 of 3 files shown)


dragonsinth (changes requested):
Hi, thanks for the contribution but it's going to be more complicated than what's in this PR.  It's actually quite complex to keep things in sync correctly.  Notice this doc on the request object:

```
     * There are three representations of the request to keep in sync: (1) the
     * raw JSON, (2) the request object, and (3) the form which contains a
     * hierarchy of inputs. There are only two ways that the request can be
     * modified: by editing the raw JSON or interacting with inputs on the form.
```

Adding a raw header input would need to be equally sophisticated.

PR #295, opened 2024-01-12
Remove -plaintext by default

rogerhu:
I'm sure there's a way to pass in this option, so happy to discuss. =)


jhump:
These arguments should really be plumbed through from the server, so that the same relevant command-line args provided to `grpcui` are also present in the `grpcurl` command line.

halkeye:
https://github.com/fullstorydev/grpcui/pull/400

PR #306, opened 2024-03-12
chore!: update dependencies

Kryvchun:
Breaking change: breaks go 1.18.

Update all dependencies to resolve broken packages.


dragonsinth:
I think a lot of these updates have now happened

PR #314, opened 2024-04-18
Support `grpc.reflection.v1.ServerReflection`

y-yagi:
Currently, `grpc.reflection.v1.ServerReflection` is shown in the list of `Service Name`. This PR removes it as well as `grpc.reflection.v1alpha.ServerReflection`.

Also, change to use `v1` protocol if we can use it.


jhump (approved):
FWIW, this looks good to me.

dragonsinth (approved):
thanks for the contribution!

PR #315, opened 2024-04-26
Added middlewares support

eyudkin:
Hi!
I found that I can not specify any dynamic callbacks like middlewares/interceptors if I use standalone server.
I think that possibility to do that is a good option so I added a support for that.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -23,8 +23,9 @@ import (
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 
-	"github.com/fullstorydev/grpcui/internal"
 	"github.com/fullstorydev/grpcurl"
+
+	"github.com/fullstorydev/grpcui/internal"
 )
 
 // RPCInvokeHandler returns an HTTP handler that can be used to invoke RPCs. The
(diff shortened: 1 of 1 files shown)


dragonsinth:
this seems reasonable, @jhump what do you think?

handlers.go
```diff
@@ -23,8 +23,9 @@ import (
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 
-	"github.com/fullstorydev/grpcui/internal"
 	"github.com/fullstorydev/grpcurl"
+
+	"github.com/fullstorydev/grpcui/internal"
```
dragonsinth, line 28: Is this separation because grpcui is this module wherease grpcurl is a remote module?

handlers.go
```diff
@@ -96,7 +119,25 @@ func RPCInvokeHandlerWithOptions(ch grpc.ClientConnInterface, descs []*desc.Meth
...
+					return invokeRPC(ctx, method, ch, descSource, r.Header, r.Body, &options)
+				}
+				for i := len(options.Middlewares) - 1; i > 0; i-- {
+					mw := options.Middlewares[i]
+					c2 := call
+					call = func(ctx context.Context, req *RPCRequest) (*RPCResult, error) {
+						return mw(ctx, req, c2)
+					}
+				}
+				results, err := call(r.Context(), req)
```
dragonsinth, line 140: There's something slightly off for me about the local variable names here, but I can't put my finger on it.  Lemme play with this a minute.

dragonsinth, line 140: How about this?

```suggestion
				handler := func(ctx context.Context, req *RPCRequest) (*RPCResult, error) {
					return invokeRPC(ctx, method, ch, descSource, r.Header, r.Body, &options)
				}
				for i := len(options.Middlewares) - 1; i > 0; i-- {
					previousMiddleware := options.Middlewares[i]
					nextHandler := handler
					handler = func(ctx context.Context, req *RPCRequest) (*RPCResult, error) {
						return previousMiddleware(ctx, req, nextHandler)
					}
				}
				results, err := handler(r.Context(), req)
```

jhump (commented):
This seems a little odd to invent a new way to do this and introduce new exported types for this. Why can't you use interceptors on the `grpc.ClientConnInterface` instead?

If you want to decorate an existing such interface (i.e. adding _more_ middleware to an existing client, instead of configuring the client with your middleware), you could use [`grpchan.InterceptClientConn`](https://pkg.go.dev/github.com/fullstorydev/grpchan#InterceptClientConn).

It's unclear why there should be a way to do this that is so specific to the UI's invocation method.

Issue #316, opened 2024-04-27
background image (grpc-bg.svg) is too large


lqs:
The background image (grpc-bg.svg) is significantly large in size (1.7MB). When embedded into servers with limited bandwidth, it consumes a significant portion of the available resources and results in slower load times.

Please consider replacing the existing image with a simpler pattern or opt for a plain color background.

dragonsinth:
You right. https://github.com/fullstorydev/grpcui/pull/317

PR #317, opened 2024-04-27
replace background image (grpc-bg.svg) with a simple gradient

dragonsinth:
https://github.com/fullstorydev/grpcui/issues/316

> The background image (grpc-bg.svg) is significantly large in size (1.7MB). When embedded into servers with limited bandwidth, it consumes a significant portion of the available resources and results in slower load times.
> 
> Please consider replacing the existing image with a simpler pattern or opt for a plain color background.


dragonsinth:
CC: @jhump what was the story on that SVG anyway?

lqs (approved):

gpassini (approved):

jhump:
A little surprising that it is so large... I just lifted it from the [grpc.io](https://grpc.io/) home page (which still uses it today). I didn't think too much about it beyond that. My bad.

dragonsinth:
> A little surprising that it is so large... I just lifted it from the [grpc.io](https://grpc.io/) home page (which still uses it today). I didn't think too much about it beyond that. My bad.

No big deal, I'm surprised anyone ever noticed!

PR #323, opened 2024-05-29
fix request file bytes update issue

nordicdyno:
file content hadn't update after file upload, only after setting focus on text field


dragonsinth:
🤷 I have no idea, @jhump I think you wrote most of this, does this pass the sniff test?

jhump:
It worked when I tested it when I originally wrote it. But maybe the quirk is browser-specific or something in newer versions of Chrome not acting the same 🤷.

Fix seems fine to me.

dragonsinth (approved):

PR #334, opened 2024-09-09
makefile updates

dragonsinth:


download_protoc.sh
```diff
@@ -31,5 +31,5 @@ if [[ "$(${PROTOC} --version 2>/dev/null)" != "libprotoc 3.${PROTOC_VERSION}" ]]
   mkdir -p .tmp/protoc
   curl -L "https://github.com/google/protobuf/releases/download/v${PROTOC_VERSION}/protoc-${PROTOC_VERSION}-${PROTOC_OS}-${PROTOC_ARCH}.zip" > .tmp/protoc/protoc.zip
   pushd ./.tmp/protoc && unzip protoc.zip && popd
+  touch -c ./.tmp/protoc/bin/protoc
```
gpassini, line 34: I'm missing the subtlety here. This command will never fail, right? So what we want to do is really change its modification time? Why?

gpassini (approved):
LGTM, but I have a question.

download_protoc.sh
dragonsinth: Unpacking protoc gives it a very old timestamp, but we reference it as a dependency in the Makefile, so we actually want to have it use whatever time it got unpacked.  Otherwise, changing protoc versions won't force a rebuild.

PR #339, opened 2024-09-26
Fix the incorrect parameter order

cuishuang:


dragonsinth (approved):
omg.. how did this survive so long... @jhump

jhump:
Lol, I guess I didn't play that much with trailers when playing around with the test server 😂

PR #348, opened 2024-11-28
Bump grpcurl dependency to v1.9.2

gpassini:
Update the version of the `grpcurl` dependency, which also bumps the minimum Go version to 1.21.


go.mod
```diff
@@ -1,9 +1,11 @@
 module github.com/fullstorydev/grpcui
 
-go 1.19
+go 1.21
+
+toolchain go1.21.3
```
dragonsinth, line 5: is this toolchain thing actually needed? i'm a little unclear on what it does and what it means to different go versions

dragonsinth (approved):

go.mod
gpassini: I'm not sure, I can search if this is needed. But this was automatically added by `go mod tidy` after I simply updated the version of the `grpcurl` dependency.

gpassini: Ok, so it seems that we don't _need_ this `toolchain` directive, because it'd be implicit by the `go` directive, which points to the same version, but since this is added by `go mod tidy`, I'm tempted to just leave it.

PR #350, opened 2024-11-29
Run full set of checks using Go 1.23

gpassini:
Some of our static check tools were too old and didn't work with Go >= 1.20.

This PR bumps the versions of those tools so they can work with the current latest Go version (1.23) and runs the full set of CI checks on that version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ ci: deps checkgofmt checkgenerate vet staticcheck ineffassign predeclared test
 
 .PHONY: deps
 deps:
-	go get -d -v -t ./...
+	go get -v -t ./...
 	go mod tidy
 
 .PHONY: updatedeps
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ ineffassign:
 
 .PHONY: predeclared
 predeclared:
-	@go install github.com/nishanths/predeclared@5f2f810c9ae6
+	@go install github.com/nishanths/predeclared@245576f9a85c96ea16c750df3887f1d827f01e9c
 	predeclared ./...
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -15,7 +15,7 @@ ci: deps checkgofmt checkgenerate vet staticcheck ineffassign predeclared test
 
 .PHONY: deps
 deps:
-	go get -d -v -t ./...
+	go get -v -t ./...
```
gpassini, line 18: The `-d` flag was deprecated, and the `go get` command was outputting this warning:
```
go: -d flag is deprecated. -d=true is a no-op
```

Makefile
```diff
@@ -78,7 +78,7 @@ ineffassign:
 
 .PHONY: predeclared
 predeclared:
-	@go install github.com/nishanths/predeclared@5f2f810c9ae6
+	@go install github.com/nishanths/predeclared@245576f9a85c96ea16c750df3887f1d827f01e9c
```
gpassini, line 81: Commit: https://github.com/nishanths/predeclared/commit/245576f9a85c96ea16c750df3887f1d827f01e9c

Makefile
```diff
@@ -68,7 +68,7 @@ vet:
 
 .PHONY: staticcheck
 staticcheck:
-	@go install honnef.co/go/tools/cmd/staticcheck@v0.4.3
+	@go install honnef.co/go/tools/cmd/staticcheck@v0.5.1
```
gpassini, line 71: Using the latest release: https://github.com/dominikh/go-tools/releases/tag/2024.1.1

dragonsinth (approved):
sweet, ty

PR #354, opened 2024-12-17
Support for any

vsrikrishna32:
1) Add support for Any resolution as a part of the request form. 
2) Fetching metadata for all messages known by the backend into the web form


vsrikrishna32:
@gpassini @dragonsinth @jhump Would you please be able to take a look when possible? Thanks.

PR #360, opened 2025-01-16
Avoid overflow in various timeouts

jhump:
Mirrors the change in https://github.com/fullstorydev/grpcurl/pull/505

### DIFF
--- a/cmd/grpcui/grpcui.go
+++ b/cmd/grpcui/grpcui.go
@@ -11,6 +11,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"math"
 	"mime"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


standalone/opts.go
```diff
@@ -46,7 +46,7 @@ func WithCSS(css []byte) HandlerOption {
 }
 
 // AddJS adds a JS file to Handler, serving the supplied contents at the URI
-// "/s/<filename>" with a Content-Type of "text/javascript; charset=UTF-8". It
+// "/s/<filename>" with a ContentType-Type of "text/javascript; charset=UTF-8". It
```
dragonsinth, line 49: huh?

jhump: Well fudge. I did a refactor-rename in Goland at some point (to rename something from Content to ContentType) and it foolishly did crap like this to a lot of the source folders in my project root. My bad, should have caught this myself before pushing. Will fix shortly.

jhump: Fixed. Sorry about that.

dragonsinth (approved):
no prob, LGTM!

Issue #363, opened 2025-02-06
There’s a problematic v1.5.0 in the Go proxy cache


lqs:
When using:
```bash
go get github.com/fullstorydev/grpcui
```

it downloads v1.5.0, which is actually older than v1.4.2.

To investigate further, I ran:
```bash
curl https://proxy.golang.org/github.com/fullstorydev/grpcui/@v/list
```

and received the following versions:
```
v1.3.0
v1.4.2
v1.0.0
v1.3.3
v1.3.1
v1.5.0  <-- problematic
v1.4.1
v1.3.2
v0.2.1
v0.1.0
v1.4.0
v1.2.0
v1.1.0
v0.2.0
```

This discrepancy is confusing for users. Could you please either remove the problematic v1.5.0 from the Go proxy cache (maybe impossible?) or release a new version (e.g., v1.5.1) to override it?

dragonsinth:
Umm... sorry but we never published a 1.5.0.  I don't know where that would be coming from, but I doubt there's anything we can do about it.

lqs:
It would be better to release the current version as v1.5.1 so that the cached v1.5.0 is overridden and users get the correct version.

jhump:
It has the same date as v1.4.1 @gpassini, any chance you may have tried to release a v1.5.0 first and then changed your mind? It's possible that the Go proxy server managed to scrape and observe the v1.5.0 tag before it was deleted.

The proper way to fix this would be to add something like the following to `go.mod` and then go ahead and release 1.4.3 with that change (in order to actually push out the retractions to the proxy and sumdb cache servers).
```
retract (
    v1.5.0 // Published accidentally when preparing 1.4.1.
    v1.4.3 // Contains retractions only.
)
```
If there are any changes worth releasing, then there's no need to also retract 1.4.3 (or maybe it's 1.5.1, if there's been anything worthy of a minor version bump).

In any event, it does mean the first real 1.5 release will have to be numbered 1.5.1. (And if you retract 1.4.3, then the next real point release would be 1.4.4.)

(Also see docs on [retraction](https://go.dev/ref/mod#go-mod-file-retract).)

gpassini:
@jhump I don't remember, but maybe...
Thanks for the docs! I'll try to retract this 1.5.0 version.

jhump:
[...]

PR #364, opened 2025-02-06
Retract v1.5.0 and v1.5.1

gpassini:
After this PR is merged, I'll tag the commit with `v1.5.1`.

Fixes https://github.com/fullstorydev/grpcui/issues/363.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,3 +34,8 @@ require (
 	google.golang.org/genproto v0.0.0-20231106174013-bbf56f31fb17 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20231106174013-bbf56f31fb17 // indirect
 )
+
+retract (
+    v1.5.0 // Published accidentally.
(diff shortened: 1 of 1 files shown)


jhump:
👍 

That's kind of funny that the docs on retractions have an example that is ascending order of versions, but apparently `go mod tidy` wants to put them in descending order... 🤷

dragonsinth (approved):
LGTM

p-linnane:
@gpassini Can you clarify if the current v1.5.1 tag is correct? There is no associated release. We want to make sure things are stable before updating in Homebrew. Thanks.

dragonsinth:
> @gpassini Can you clarify if the current v1.5.1 tag is correct? There is no associated release. We want to make sure things are stable before updating in Homebrew. Thanks.

It's not a real release. We're just trying to clear the go module cache of a weird anomaly.

p-linnane:
Great. Thanks for confirmation. We will not ship that tag then.

PR #368, opened 2025-02-25
Fix filtering services or methods to expose

n0v1:
This fixes filtering services or methods to expose using the -service or -method option.

Note that in line 878 the processed config is deleted from the passed configs map. This lets the len(configs) != 0 check in line 874 fail for all following services. This results in exposing more services than expected.


n0v1:
@dragonsinth, @gpassini, could one of you two take a look please? We're using this patch since a while now and did not notice any issues.

jhump (approved):
Makes sense to me 👍

gpassini (approved):
LGTM! Thanks

PR #369, opened 2025-03-06
User Input Form Fix

ErinBaileyFullstory:
This PR restricts the form field type to text so that it cannot be interpreted as HTML


dragonsinth (approved):
👍 sweet

PR #372, opened 2025-03-11
Bump grpcurl dependency version to v1.9.3

gpassini:
Updating the `grpcurl` version to the latest release (https://github.com/fullstorydev/grpcurl/releases/tag/v1.9.3) before releasing `grpcui` itself.


dragonsinth (approved):
nice!

PR #398, opened 2025-09-25
Add comprehensive dark mode implementation for grpcui

jainal09:
## Summary
Implements a complete dark mode feature for grpcui with a toggle switch that allows users to seamlessly switch between light and dark themes.

## Features
- **Dark mode toggle switch** with accessibility features (ARIA attributes, keyboard navigation)
- **localStorage persistence** - remembers user preference across sessions
- **System preference detection** - automatically respects `prefers-color-scheme` media query
- **Professional color scheme** inspired by GitHub's dark theme
- **Comprehensive styling** for all UI elements including:
  - Form inputs, buttons, and controls
  - Tables, tabs, and response panels  
  - Error messages and status indicators
  - jQuery UI datepicker/calendar widget
- **Accessibility compliant** with proper contrast ratios
- **Cross-browser compatibility**

## Implementation Details
- Added toggle switch UI in `webform-template.html`
- Implemented theme management JavaScript in `webform.js`
- Extended CSS with comprehensive dark mode styles in `webform-sample.css`
- Uses CSS custom properties for maintainable theming
- Follows existing code patterns and conventions

## Testing
Tested comprehensively with the KitchenSink test server to ensure all UI elements (including calendar widgets with timestamp fields) render correctly in both light and dark modes.

## Screenshots
The implementation provides a professional dark theme that matches modern UI expectations while maintaining full functionality and accessibility.

### DIFF
--- a/internal/resources/webform/webform-sample.css
+++ b/internal/resources/webform/webform-sample.css
@@ -3,6 +3,77 @@
     font-size: 1.0em;
 }
 
+#grpc-form .grpc-theme-switch {
+    display: flex;
+    justify-content: flex-end;
+    margin-bottom: 12px;
+}
+
+#grpc-form .grpc-theme-toggle {
+    display: inline-flex;
+    align-items: center;
+    gap: 10px;
+    font-size: 90%;
+    cursor: pointer;
+    user-select: none;
+}
+
+#grpc-form .grpc-theme-toggle input {
+    position: absolute;
+    width: 1px;
+    height: 1px;
+    margin: -1px;
+    padding: 0;
+    border: 0;
+    clip: rect(0 0 0 0);
+    overflow: hidden;
+}
+
+#grpc-form .grpc-theme-toggle-track {
+    position: relative;
+    width: 44px;
+    height: 24px;
+    border-radius: 999px;
+    background-color: #d0d0d0;
+    transition: background-color 0.2s ease;
+    flex-shrink: 0;
+}
+
+#grpc-form .grpc-theme-toggle-thumb {
+    position: absolute;
+    top: 3px;
+    left: 3px;
+    width: 18px;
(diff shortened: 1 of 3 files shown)


jainal09:
@gpassini  @dragonsinth  @jhump 

Hi! I know u guys might be busy, but I would much appreciate a review as per your convivence

jainal09:
@dragonsinth ,  Thank you for the feedback! I understand the concern about complexity, but
   I'd like to address a key usability issue with a pure CSS-only approach.

  The Problem:
  - Pure CSS `light-dark()` only respects system preferences with no override
  capability
  - So, If users want to use a different theme than their system setting, they
  have no way to do so. 
  - We can keep a pure css toggle without local storage but Without any persistence, users would have to manually toggle the theme
  on every page refresh, which creates a poor user experience

 So, I propose we find a middle ground approach:
  - We replace the complex toggle switch with a simple sun/moon icon button.
  - We keep minimal localStorage (just one key) for theme preference - this is
  essential UX
  - We also significantly simplify the JavaScript (remove system preference change
  listeners, complex state management)
  - use light-dark() CSS as the base with simple override classes
  - So, By default if there is no darkmode switch ever done, we still respect system preferences but allow
  users to override and remember their choice.

  The localStorage is really the minimum viable functionality needed. Modern
   web users expect their theme choice to persist, and it's just a single
  localStorage key. Without it, the feature becomes more of a frustration
  than a benefit.

  Would this simplified approach with minimal persistence be acceptable?

jainal09:
I made some changes and this is how it will look like if you approve it
image
image

This looks minimal and not too overkill

jainal09:
@dragonsinth  done. Removed all custom js and UI button and implemented a pure CSS based darkmode that detects from system settings and applies the dark mode theme!

PR #400, opened 2025-10-07
Un-hardcode -plaintext in grpcurl text box. Use the same flags as ui

halkeye:
An alternative method could be populating a header with this data, would remove the global array

### DIFF
--- a/standalone/standalone.go
+++ b/standalone/standalone.go
@@ -48,7 +48,7 @@ func Handler(ch grpcdynamic.Channel, target string, methods []*desc.MethodDescri
 		indexTmpl:      defaultIndexTemplate,
 		css:            grpcui.WebFormSampleCSS(),
 		cssPublic:      true,
-		gRPCurlOptions: []string{"-plaintext"},
+		gRPCurlOptions: nil,
 	}
 	for _, o := range opts {
 		o.apply(uiOpts)


internal/resources/webform/webform.js
```diff
@@ -2133,7 +2134,7 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
             }
         }
 
-        gRPCurlTextArea.html(`<div>grpcurl -plaintext${metadataStr} -d '${requestDataJson}' ${window.target} ${service}.${method}</div>`);
```
dragonsinth, line 2136: What happened to `${metadataStr}`?

cmd/grpcui/grpcui.go
```diff
@@ -351,6 +361,18 @@ func main() {
 	flags.Usage = usage
 	flags.Parse(os.Args[1:])
 
+	gRPCOptions := []string{}
+	for _, flagName := range grpcCurlFlags {
+		f := flags.Lookup(flagName)
+		if f.Value.String() != f.DefValue {
+			if f.Value.String() == "true" {
```
dragonsinth, line 368:
```suggestion
			if getter, ok := f.Value.(flag.Getter); ok && getter.Get() == true {
```
Otherwise you could break on a string value that happens to be the string true.

standalone/opts.go
```diff
@@ -15,6 +15,9 @@ type WebFormContainerTemplateData struct {
 	// Target is the name of the machine we are making requests to (for display purposes).
 	Target string
 
+	// GRPCurlOptions is the list of options needed to be passed through to grpcurl, such as -key, or -plaintext converted to raw json
+	GRPCurlOptionsJSON template.JS
```
dragonsinth, line 19:
```suggestion
	GRPCurlOptions string
```

standalone/standalone.go
```diff
@@ -155,6 +157,12 @@ func getIndexContents(tmpl *template.Template, target string, webFormHTML []byte
 		WebFormContents: template.HTML(webFormHTML),
 		AddlResources:   addlHTML,
```
dragonsinth, line 158: GRPC
```suggestion
		AddlResources:   addlHTML,
		GRPCurlOptions:  strings.Join(gRPCOptions, " "),
```

internal/resources/webform/webform.js
```diff
@@ -2126,11 +2127,11 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
             var name = $(cells[0]).val();
             var val = $(cells[1]).val();
             if (name !== "") {
-                metadataStr += ` -H "${name}: ${val}"`;
+                opts.push(`-H "${name}: ${val}"`);
             }
         }
 
-        gRPCurlTextArea.html(`<div>grpcurl -plaintext${metadataStr} -d '${requestDataJson}' ${window.target} ${service}.${method}</div>`);
+        gRPCurlTextArea.html(`<div>grpcurl ${opts.join(" ")} -d '${requestDataJson}' ${window.target} ${service}.${method}</div>`);
```
dragonsinth, line 2134:
```suggestion
        gRPCurlTextArea.html(`<div>grpcurl ${opts}${metadataStr} -d '${requestDataJson}' ${window.target} ${service}.${method}</div>`);
```
and revert the other changes

standalone/standalone.go
```diff
@@ -293,7 +301,7 @@ func computeETag(contents []byte) string {
 // and methods supported by the server, and constructs a handler to serve the UI.
 //
 // The handler has the same properties as the one returned by Handler.
-func HandlerViaReflection(ctx context.Context, cc grpc.ClientConnInterface, target string, opts ...HandlerOption) (http.Handler, error) {
+func HandlerViaReflection(ctx context.Context, cc grpc.ClientConnInterface, target string, gRPCOptions []string, opts ...HandlerOption) (http.Handler, error) {
```
dragonsinth, line 304: This seems like a breaking API change.. maybe we need a new HandlerOption here?

webform.go
```diff
@@ -76,20 +77,21 @@ type WebFormOptions struct {
 // accepts an additional argument, options. This can be used to toggle the JS
 // code into debug logging and can also be used to define the set of metadata to
 // show in the web form by default (empty if unspecified).
-func WebFormContentsWithOptions(invokeURI, metadataURI string, target string, descs []*desc.MethodDescriptor, opts WebFormOptions) []byte {
+func WebFormContentsWithOptions(invokeURI, metadataURI string, target string, grpcOptions []string, descs []*desc.MethodDescriptor, opts WebFormOptions) []byte {
```
dragonsinth, line 80: same with these... seems breaking?

standalone/standalone.go
dragonsinth: I think nil is fine and defaulting to -plaintext on the client is fine

halkeye:
I see whats going on more now. My bad. I was hacking something quickly while working

image

I don't have a grpc server with tls outside of work, but i tried it out with the grpc reflection example and everything seems to work good.

dragonsinth (approved):
LGTM just a couple of tiny nits; this looks great.

internal/resources/webform/webform.js
```diff
@@ -2113,24 +2113,25 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
...
-            var val = $(cells[1]).val();
+            const name = $(cells[0]).val();
+            const val = $(cells[1]).val();
             if (name !== "") {
                 metadataStr += ` -H "${name}: ${val}"`;
             }
         }
 
-        gRPCurlTextArea.html(`<div>grpcurl -plaintext${metadataStr} -d '${requestDataJson}' ${window.target} ${service}.${method}</div>`);
+        gRPCurlTextArea.html(`<div>grpcurl ${window.gRPCurlOptions}${metadataStr} -d '${requestDataJson}' ${window.target} ${service}.${method}</div>`);
```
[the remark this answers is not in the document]
dragonsinth: Huh, I wonder why we even create that div to begin with instead of just setting innerText on the <pre> tag...

PR #421, opened 2026-03-21
feat(webform): add service/method comboboxes with URL and localStorage restore

isqua:
https://github.com/user-attachments/assets/6f10876e-9dc7-4591-a8dc-bb9b418747e1

Hello, thank you for maintaining this awesome project!

I recently joined a team, and we have a lot of services. So to find a service I want, I had to read the list of options in the service selector to find and click the necessary one. So I added a few enhancements:

1. Comboboxes for "Service name" and "Method name". You now can type service name, e.g. `auth`, and then select (using mouse or keyboard) from the filtered list of options.
2. Persistance for "Service name" and "Method name". So when you close the grpcui, then after a while you come back, the same service and method names are selected (that were there the last time)
3. Taking "Service name" and "Method name" from URL. So let's say you're have a documentation for your services, there you can put links like `my-grpcui-instance.com/?serviceName=com.company.OrderService&methodName=GetOrder`, and when users click such a link, they go to the GRPC UI with the preselected service and method. Startup precedence is: query params -> localStorage -> first available service/method

### DIFF
--- a/internal/resources/webform/webform.js
+++ b/internal/resources/webform/webform.js
@@ -26,7 +26,6 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
             methodList.val(methods[0]);
         }
         syncComboboxInput("#grpc-method");
-        persistSelectionState();
         // implicit selection of first element does not
         // generate a change event, so we have to do this
         formMethodSelected(callback);
@@ -2642,9 +2641,6 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
 
     const setupCombobox = (selectSelector, inputId) => {
         const selectEl = $(selectSelector);
-        if (selectEl.length === 0 || selectEl.data("comboboxReady")) {
-            return;
-        }
 
         const inputEl = $("<input>", {
             id: inputId,
@@ -2654,16 +2650,17 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
             "aria-haspopup": "listbox",
         }).insertAfter(selectEl);
 
-        const readSource = () => getSelectOptions(selectEl);
         inputEl.val(selectEl.val() || "");
 
         inputEl.autocomplete({
             minLength: 0,
             delay: 0,
             source: function(req, resp) {
-                const options = readSource();
+                const options = getSelectOptions(selectEl);
+                // When user types an exact option value, put it first in the dropdown
+                // so it's pre-selected, then show remaining options for easy browsing
                 const exactMatchIndex = options.findIndex(item => item === req.term);
-                if (exactMatchIndex >= 0) {
+                if (exactMatchIndex !== -1) {
                     const exactMatch = options[exactMatchIndex];
                     const reordered = [exactMatch].concat(options.filter((item, index) => index !== exactMatchIndex));
                     resp(reordered);
@@ -2700,8 +2697,8 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
         selectEl.on("change.comboboxSync", function() {
             syncComboboxInput(selectSelector);
         });
-        selectEl.hide();
         inputEl.width(selectEl.outerWidth());
+        selectEl.hide();
         selectEl.data("comboboxInput", inputEl);
         selectEl.data("comboboxReady", true);
     };
@@ -2743,6 +2740,13 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
         }
         if (hasMethodForService(service, method)) {
             setStorageItem(methodStorageKey, method);
+        }
+        // Update URL query params to make the current selection shareable
+        if (hasService(service) && hasMethodForService(service, method)) {
+            const url = new URL(window.location);
+            url.searchParams.set("serviceName", service);
+            url.searchParams.set("methodName", method);
+            window.history.replaceState(null, "", url.toString());
         }
     };
 
--- a/internal/resources/webform/webform.js
+++ b/internal/resources/webform/webform.js
@@ -3,7 +3,7 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
     var descriptionsShown = false;
     var requestForm = $("#grpc-request-form");
 
-    function formServiceSelected(callback) {
+    function formServiceSelected(callback, preferredMethod) {
         var svcName = $("#grpc-service").val();
         var svcDesc = svcDescs[svcName];
         var methods = services[svcName];
@@ -17,10 +17,16 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
         var methodList = $("#grpc-method");
         methodList.empty();
         for (var i = 0; i < methods.length; i++) {
-            m = methods[i];
+            var m = methods[i];
             methodList.append($("<option>", {value: m, text: m}));
         }
-        $("#grpc-method option:first-of-type").select();
+        if (preferredMethod && methods.includes(preferredMethod)) {
+            methodList.val(preferredMethod);
+        } else if (methods.length > 0) {
+            methodList.val(methods[0]);
+        }
+        syncComboboxInput("#grpc-method");
+        persistSelectionState();
         // implicit selection of first element does not
         // generate a change event, so we have to do this
         formMethodSelected(callback);
@@ -41,6 +47,7 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
         $.ajax(metadataURI + "?method=" + fullMethod)
             .done(function(data) {
                 buildRequestForm(data);
+                persistSelectionState();
                 callback?.();
             })
             .fail(function(data, status) {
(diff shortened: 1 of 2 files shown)


dragonsinth:
This is super cool! What a great idea. Will try to review when I get a chance, tagging @franklinlindemberg and @gpassini in case one of them wants to take the review first.

franklinlindemberg:
> This is super cool! What a great idea. Will try to review when I get a chance, tagging @franklinlindemberg and @gpassini in case one of them wants to take the review first.

I really liked this! Definitely an welcome change!

isqua:
Hey guys, any news? What can I do to speed up the code review?

franklinlindemberg:
> Hey guys, any news? What can I do to speed up the code review?

@isqua sorry for the delay, Im reviewing it this week

internal/resources/webform/webform.js
```diff
@@ -2575,6 +2582,169 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
...
+
+        inputEl.on("focus", function() {
+            $(this).autocomplete("search", $(this).val());
+        });
+
+        selectEl.on("change.comboboxSync", function() {
+            syncComboboxInput(selectSelector);
+        });
+        selectEl.hide();
+        inputEl.width(selectEl.outerWidth());
```
franklinlindemberg, line 2700: nit: could we swap these 2 lines? Im afraid in some browsers it might not work as we expect (we might not be able to get the right width if the element is hidden)

internal/resources/webform/webform.js
```diff
@@ -2575,6 +2582,169 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
...
+        selectEl.data("comboboxInput", inputEl);
+        selectEl.data("comboboxReady", true);
+    };
+
+    const resolveInitialSelection = () => {
+        const params = new URLSearchParams(window.location.search || "");
+        const serviceFromQuery = params.get("serviceName");
```
franklinlindemberg, line 2708: > Taking "Service name" and "Method name" from URL. So let's say you're have a documentation for your services, there you can put links like my-grpcui-instance.com/?serviceName=com.company.OrderService&methodName=GetOrder, and when users click such a link, they go to the GRPC UI with the preselected service and method. Startup precedence is: query params -> localStorage -> first available service/method

@isqua I noticed we are never updating the url with the service/method names. Did you mean to do that? Otherwise Im not sure if the ability to read from the url will be useful (unless people add those query params manually

internal/resources/webform/webform.js
```diff
@@ -17,10 +17,16 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
...
             methodList.append($("<option>", {value: m, text: m}));
         }
-        $("#grpc-method option:first-of-type").select();
+        if (preferredMethod && methods.includes(preferredMethod)) {
+            methodList.val(preferredMethod);
+        } else if (methods.length > 0) {
+            methodList.val(methods[0]);
+        }
+        syncComboboxInput("#grpc-method");
+        persistSelectionState();
```
franklinlindemberg: what do you think if we remove this call to persistSelectionState and leave just the one on line 50?

internal/resources/webform/webform.js
```diff
@@ -2575,6 +2582,169 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
...
+        const selectEl = $(selectSelector);
+        const inputEl = selectEl.data("comboboxInput");
+        if (inputEl && inputEl.length) {
+            inputEl.val(selectEl.val() || "");
+        }
+    };
+
+    const setupCombobox = (selectSelector, inputId) => {
+        const selectEl = $(selectSelector);
+        if (selectEl.length === 0 || selectEl.data("comboboxReady")) {
```
franklinlindemberg: is it possible, for the same selectSelector, to run setupCombobox multiple times?

internal/resources/webform/webform.js
```diff
@@ -2575,6 +2582,169 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
...
+
+        const inputEl = $("<input>", {
+            id: inputId,
+            type: "text",
+            class: "grpc-combobox-input ui-widget ui-widget-content ui-state-default ui-corner-left",
+            "aria-autocomplete": "list",
+            "aria-haspopup": "listbox",
+        }).insertAfter(selectEl);
+
+        const readSource = () => getSelectOptions(selectEl);
```
franklinlindemberg: any reason for this to be assigned as a function instead of just storing the select options here? I believe the options are not expected to change, right?

internal/resources/webform/webform.js
```diff
@@ -2575,6 +2582,169 @@ window.initGRPCForm = function(services, svcDescs, mtdDescs, invokeURI, metadata
...
+        inputEl.val(selectEl.val() || "");
+
+        inputEl.autocomplete({
+            minLength: 0,
+            delay: 0,
+            source: function(req, resp) {
+                const options = readSource();
+                const exactMatchIndex = options.findIndex(item => item === req.term);
+                if (exactMatchIndex >= 0) {
```
franklinlindemberg: what is special about exact match? Im trying to understand this if block but Im not sure I get it (maybe a comment would help).

Should we expect duplicates (exactMatchIndex > 1)?

internal/resources/webform/webform.js
isqua: Fixed

internal/resources/webform/webform.js
isqua: I was building it with some generated docs in mind. But I like the idea of manually copying the URL as well, so I added the URL synchronization as you suggested

internal/resources/webform/webform.js
isqua: makes sense

internal/resources/webform/webform.js
isqua: Well, we call it only twice, so it should be safe to remove the extra guard

internal/resources/webform/webform.js
isqua: fixed

internal/resources/webform/webform.js
isqua: It is not about the count, it is an index of found element.

What problem does this code solve:

1. While the user is typing, we only show matching options, so the more the user has typed, the fewer items are visible in the dropdown.
2. So, when an option is selected, this is the input value. Therefore, we should show only one matching option in the dropdown.
3. But it makes the UX worse, since the user doesn't see any other available options.
4. To address that, it is better to show all the options. But moving the selected one to the top.

To make the code less surprising, I changed the condition to `exactMatchIndex !== -1`. When a person with javascript experience sees it, they most likely will get it as "if an index found", since `-1` is a common pattern for not-found-index. We will do a little extra job if the selected option has index 0, but the code is more usual.

Does it make sense?

isqua:
@franklinlindemberg thank you for the review! I addressed your comments, please have a look when you have time

franklinlindemberg:
@isqua That looks great!

While testing I noticed a small UI issue when we expand the proto details. Do you think its possible to fix it?
image

internal/resources/webform/webform.js
franklinlindemberg: I think we call it once for service and once for method, right? yeah, I think its fine to remove this extra guard

internal/resources/webform/webform.js
franklinlindemberg: yeah, I think it will be great to share the url with other team members for example, thanks for updating it

internal/resources/webform/webform.js
franklinlindemberg: It does, thanks for explaining. I dont have a strong opinion on the UI, initially I thought that we could just show a single item when it matched but I agree that the user could think there is no more options to select. Its just a bit weird that while not on full match you have less items being shown and then all of a sudden we have everything back (but we can argue that until a match is found its still a work in progress search).

In any case, Im fine moving forward with this, we can iterate back here if needed in the future

internal/resources/webform/webform.js
isqua: Yep, it's removed

isqua:
@franklinlindemberg do you mean the `<<` button floating over the code snippet? I see it in the master branch as well, so it doesn't seem related to the combobox changes. However, it can be fixed with some CSS

franklinlindemberg:
> @franklinlindemberg do you mean the `<<` button floating over the code snippet? I see it in the master branch as well, so it doesn't seem related to the combobox changes. However, it can be fixed with some CSS

I think after we resized the button now is appearing over the text area, for reference thats what it looks like here (current code)
image

let me know what you think

isqua:
Ah, interesting, it looks different for me in the master branch (specifically in af863d3fa9a15dab8c0de1fa5bdb71822136b0e3) (macOS, both Chrome and Safari)

Screenshot_2026-04-21_at_23_00_33

But I would expect the button to have gray background and borders on any OS in any browser. The button has following styles with white background and `border: none`:

https://github.com/fullstorydev/grpcui/blob/af863d3fa9a15dab8c0de1fa5bdb71822136b0e3/internal/resources/webform/webform-sample.css#L30-L39

But later in this file there's a selector with the same specificity (weight), that overrides some rules, inc. background and border:

https://github.com/fullstorydev/grpcui/blob/af863d3fa9a15dab8c0de1fa5bdb71822136b0e3/internal/resources/webform/webform-sample.css#L176-L185

Could you explain how exactly do you test, do you have a CSS minifier or something that can affect the order of rules? What commit are you checking at the main branch?

isqua:
@franklinlindemberg I reordered the CSS rules, it works for me now. Could you please check if it works for you? But if you have something that can resort CSS, it's still fragile.

franklinlindemberg (approved):
Thanks for working on this, it looks great!

isqua:
@franklinlindemberg thank you for merging! When do you think you plan to release it? It's been a year since [the last tag](https://github.com/fullstorydev/grpcui/releases/tag/v1.4.3) 😅

franklinlindemberg:
> @franklinlindemberg thank you for merging! When do you think you plan to release it? It's been a year since [the last tag](https://github.com/fullstorydev/grpcui/releases/tag/v1.4.3) 😅

Hey @isqua I was planning to let it bake for some more time, to check if there is any issue with it. Recently we found out that it broke the dark mode.

The way Im integrating right now is installing pointing to the last commit in master, would that work for you?

i believe i could do a release by the end of the month or in june

isqua:
Thank you, we'll be waiting then!

isqua:
Hi @franklinlindemberg! Any chance of a new release in the near future?

franklinlindemberg:
> Hi @franklinlindemberg! Any chance of a new release in the near future?

yeah! Ill do this week, Ill just try to fix something before that

franklinlindemberg:
> Hi @franklinlindemberg! Any chance of a new release in the near future?

@isqua  release v1.5.2 is out

isqua:
@franklinlindemberg Awesome, thank you!

PR #422, opened 2026-03-30
Upgrade google.golang.org/grpc from v1.66.2 to v1.79.3

olegbonar:
Migrate deprecated grpc.DialContext + grpc.WithBlock to grpc.NewClient with explicit connection readiness polling via connectivity state, and use passthrough resolver scheme to preserve existing address behavior.

**This is an AI-generated code change**


dragonsinth:
Are code updates required for this version bump, or are the changes separable?

olegbonar:
> Are code updates required for this version bump, or are the changes separable?

Those are necessary since the library api has changed

dragonsinth:
This seems like quite a big change from what's there. I was looking at grpcurl for comparison and I don't see anything quite like this in that code base.  generally grpcui should probably take direction from grpcurl

olegbonar:
@dragonsinth but grpcurl is still depending on google.golang.org/grpc v1.66.2

dragonsinth:
> @dragonsinth but grpcurl is still depending on google.golang.org/grpc v1.66.2

Ahh, that would explain it. Well either way we should probably fix it there first and then copy.  In fact, grpcui already depends on grpcurl-- can we just call `grpcurl.BlockingDial` from grpcui, and then shunt any fixes over to that project?

CC: @jhump

olegbonar:
https://github.com/fullstorydev/grpcurl/pull/556

Issue #438, opened 2026-08-07
[Feature Request] Forward specified headers between GRPC backend and web UI


gbargoud:
I want to set the login session for an RPC server in a cookie which would require forwarding the Set-Cookie and Cookie headers from the GRPC backend to the web UI.

If I were to make a pull request implementing this, would it likely be accepted? 

## Some Implementation options

1. Rename `PreserveHeaders` in [InvokeOptions](https://github.com/fullstorydev/grpcui/blob/master/handlers.go#L50) to `PreserveRequestHeaders` and add a similar `PreserveResponseHeaders` field (which can be set in the same way)
2. Add a `PreserveResponseHeaders` without renaming `PreserveHeaders`
    * Leaning for this option since it has minimal effect on existing use cases
4. Apply the existing `PreserveHeaders` to both request and response headers

gbargoud:
Another option would be a "preserve cookies" flag in which specific `Set-Cookie` and `Cookie` headers are set to be forwarded between the web UI and gRPC backend.

gbargoud:
Implemented option 2 from the original comment in https://github.com/gbargoud/grpcui/tree/forward_headers and tested that it forwards the cookies for the appropriate sessions to my backend.

I might also implement the suggestion in my other comment in a new branch to see which one is better.

PR #442, opened 2026-08-28
Preserve specified cookies between browser and gRPC UI

gbargoud:
Adds a new flag to forward specified cookies between the browser and the gRPC UI.

This makes auth sessions easier to implement for a gRPC server by using the Set-Cookie header in the response and adding the appropriate preserve-cookie flag when launching grpcui.

See https://github.com/fullstorydev/grpcui/issues/438

Tested against [an RPC server that uses cookies for auth sessions](http://codeberg.org/gbargoud/hotch):

```
go build cmd/grpcui/grpcui.go && ./grpcui -preserve-cookie=session_id -preserve-cookie=user_id -plaintext 127.0.0.1:8000
```

---

> [!NOTE]
> **Medium Risk**
> Opt-in cookie forwarding touches auth/session paths; misconfiguration could leak or overwrite cookies, though names are explicitly allowlisted and full Cookie preservation via preserve-header still takes precedence.
> 
> **Overview**
> Adds **selective cookie forwarding** so gRPC UI can carry cookie-based auth between the browser and the backend, similar to the existing preserve-header behavior but scoped by cookie name.
> 
> The **`grpcui` CLI** gains a repeatable **`-preserve-cookie`** flag; values are passed through **`standalone.PreserveCookies`** into **`InvokeOptions.PreserveCookies`**. On RPC invoke, matching cookies from the browser **`Cookie`** header are attached as gRPC metadata (unless **`-preserve-header Cookie`** already forwards the full header). After the RPC, **`Set-Cookie`** values from response metadata whose names are in the allowlist are written on the HTTP response so the browser can store session cookies from the server.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed49a6f919f581a154ebb21b04f818afe1032415. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dragonsinth:
I think I like the version from https://github.com/fullstorydev/grpcui/compare/master...gbargoud:grpcui:forward_headers that uses PreserveResponseHeaders better maybe? @jhump do you have anything thoughts on this? Seems parallel to existing PreserveHeader (request)

